PR #9, opened 2015-02-11
Wire ts tests

mhegazy:
This change splits any system dependency (e.g. node FileSystem) from the core session code, to allow for testability. All system-dependent code is in server.ts which is the command-line driver, 

I have also added a simple JS client. currently this does not support all operations just a small subset. more to come.


billti:
This pull request is a little old now and cannot be merged cleanly. Is this still in progress, or is this obsolete?  Thanks.

mhegazy:
These files have been ported a while back. closing.

PR #110, opened 2015-04-03
Avoid unnecessary file opening before closing files

### LINKED ISSUE #104: ST3 is slow to shut down when lots of TS files are open
This seems to be recent - when I close Sublime and I have lots of TS files open (20 +), it takes 10 or 15 seconds before the process is gone from Task Manager. If I try to open ST again during that time, nothing happens. When I only have lots of non-TS files open, it shuts down immediately.

zhengbli:
Fix #104

### DIFF
--- a/TypeScript.py
+++ b/TypeScript.py
@@ -433,6 +433,7 @@ class TypeScriptListener(sublime_plugin.EventListener):
         self.errRefreshRequested = False
         self.changeFocus = False
         self.mod = False
+        self.batch_close_queue = []
 
     def getInfo(self, view):
         info = None
@@ -686,7 +687,11 @@ class TypeScriptListener(sublime_plugin.EventListener):
                # make sure we know the latest state of the file
                reloadBuffer(view, info.clientInfo)
                # notify the server that the file is closed
-               cli.service.close(view.file_name())
+               self.batch_close_queue.append(view.name())
+               if len(self.fileMap) <= 1 or len(self.batch_close_queue) = 5:
+                    for fname in self.batch_close_queue:
+                        cli.service.close(fname)
+                    self.batch_close_queue = []
           
 
     # called by Sublime when the cursor moves (or when text is selected)


TypeScript.py
```diff
@@ -686,7 +687,11 @@ def on_close(self, view):
                # make sure we know the latest state of the file
                reloadBuffer(view, info.clientInfo)
                # notify the server that the file is closed
-               cli.service.close(view.file_name())
+               self.batch_close_queue.append(view.name())
+               if len(self.fileMap) <= 1 or len(self.batch_close_queue) = 5:
```
DanielRosenwasser: You mean `==`?

DanielRosenwasser: I mean instead of `=`

zhengbli: Thanks for pointing out. And it turns out I need to batch the process in the server side also, so much more code changes needed to actually fix this. Will commit later.

zhengbli:
So, after some more investigation, I believe the most possible reason why the editor hanged when batch closing happens is unnecessary file opening.

The logging shows when closing a file, the following event listeners are triggered in order:

1. ``on_activated`` (only triggered when using "close all" command)
2. ``on_close``

And both methods call ``getInfo``, which tries to open the file on the server. Therefore if any of the files to be closed has never been opened on server before, the plugin will try to do that right before closing it, which involves creating a temp project system for each file and is expensive and unnecessary. And to make things worse, although sublime appears to remember all the files in a window after you close and reopen the window, it doesn't load them until you activate the view. So it is not hard to explain why the hang happens a lot. 

So my proposal tries to solve the two common cases of batch closing:
- via ``close all`` window command:
In this case sublime will trigger ``on_window_command`` event listener. So I set a global flag in this listener to notify the following closings that this is a batch closing, so don't try to open the files;

- via clicking on the red cross button on the window's title bar:
In this case ``on_window_command`` event listener won't be triggered. However, sublime will not trigger ``on_activated`` on each view either. Therefore any unopened file will remain unopened when ``on_close`` is called. And I can just avoid unopened files in ``on_close``.

Please comment if there are anything wrong in my attempt to solve this problem, or if there are better solutions. Thanks!

zhengbli:
For some reason the pull request wouldn't update my commits..

Issue #92, opened 2015-03-27
Can only see the first call signature in a set of overloads


billti:
For a function with a set of overloaded signatures, I am only able to view the first overload (see below, unable to cycle through other overloads).  Is this a TODO, a bug, or a command/key stroke I am unable to locate?

![screen shot 2015-03-26 at 10 46 14 pm](https://cloud.githubusercontent.com/assets/993909/6863259/37d00b18-d40b-11e4-9e20-873ed1313a1f.png)

billti:
See the write up in the pull request.  This seems to work quite well when there are a lot of overloads (e.g. addEventListener), as the fuzzy searching helps you narrow down quickly.  The args-as-snippets behavior is quite handy too.

billti:
This is merged in. We'll continue working on ideas to expose it from other areas too (i.e. a link on the tooltip).

PR #112, opened 2015-04-04
Show signature overloads (and completions) in the Quick Panel

billti:
This is a work in progress, just putting it up here for feedback.  See #92 for more details.

Effectively, after you type the opening paren of a function call, if you hit the question mark '?', then a list of overloads will pop up in the quick panel (see below for when cursor was in the jQuery `attr` function on line 10).  If you select one of these, it will populate the arguments as snippets, so you can tab over each param and replace it with the desired args.  (It may look better if the param list is wrapped in parens, and add the return type..., or maybe even repeat the full sig with name as well - thoughts?)

screen shot 2015-04-04 at 2 32 55 am

Note I added a similar mechanism for the member completion list too (press '?' after the dot operator) in case people prefer the look and feel of the Quick Panel, but this may not be needed with completions working here anyway.

The completion panel is also available via [ctrl-t, ctrl-c], the the signature overloads panel by [ctrl-t, ctrl-o].

Still need to work on tidying up some scenarios, but happy for feedback before iterating further.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
-TypeScript.tmLanguage.cache
-FindRefs.hidden-tmLanguage.cache
+*.cache
 .tmpbuf
 .log*
 built/*
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -87,6 +87,20 @@
             { "key": "selector", "operator": "equal", "operand": "source.ts" }
         ]
     },
+    {
+        "keys": [ "ctrl+t", "ctrl+c" ],
+        "command": "typescript_completion_panel",
+        "context": [
+            { "key": "selector", "operator": "equal", "operand": "source.ts" }
+        ]
+    },
+    {
+        "keys": [ "ctrl+t", "ctrl+o" ],
+        "command": "typescript_signature_panel",
+        "context": [
+            { "key": "selector", "operator": "equal", "operand": "source.ts" }
+        ]
+    },
     // In case when auto match is enabled, only format if not within {}
     {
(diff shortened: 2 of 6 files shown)


steveluc:
:+1:   
Like this.  It's a nice way to select overloads.  Great to have the signature help hooked up.  

I propose we combine it with method tip popup by putting a link on the method tip to open the overload selector if there are overloads.   I like the idea of putting in the snippets once the overload is selected.  I also like the idea of just putting in the snippet in the case if a single overload when you type '?' after the '('  We'll need to make sure we can update the method tip when people are tabbing, because people may also want the method tip up even when the snippets are present, since when you type over the text field for a parameter, the parameter type information will disappear.  

Don't know how many people will want to use the completions one, but we can try it.

billti:
Cleaned up a little.  It now displays the full signature in the panel, as this is the clearest (and sometimes the return type is of interest).  It appears as below now for the example shown above.

screen shot 2015-04-05 at 6 39 08 pm

I removed the Completion Panel.  You're right, this may not be very useful to people with completions working anyway, and I'd rather keep feature set and code as simple as possible (but no simpler :smile:)

I tried the 'automatically insert if only one overload', but the problem there is, sometimes I just want to see the panel to see the signature (especially on Sublime Text 2 with no tooltips), so automatically pasting in the snippet is a pain to then undo every time.  If there's only one signature, it's only one more key press (`enter`) after the list appears to insert it.

I'll work on the tooltip popup link after I merge this (pending a thumbs up for the new changes).

steveluc:
:+1:

PR #115, opened 2015-04-05
Avoid unnecessary file opening before closing files

### LINKED ISSUE #104: ST3 is slow to shut down when lots of TS files are open

zhengbli:
Reopen pull request because for some reason github won't update the last one. This is to fix #104.

### DIFF
--- a/TypeScript.py
+++ b/TypeScript.py
@@ -648,12 +648,17 @@ class TypeScriptListener(sublime_plugin.EventListener):
     def on_load(self, view):
         logger.view_debug(view, "enter on_load")
         clientInfo = cli.getOrAddFile(view.file_name())
+
+        # reset the "close_all" flag when open new files
+        if self.about_to_close_all:
+            self.about_to_close_all = False
+
         print("loaded " + view.file_name())
         if clientInfo and clientInfo.renameOnLoad:
             view.run_command('typescript_delayed_rename_file',
                              { "locsName" : clientInfo.renameOnLoad })
             clientInfo.renameOnLoad = None
-        logger.view_debug(view, "enter on_load")
+        logger.view_debug(view, "exit on_load")
 
     def on_window_command(self, window, command_name, args):
         # logger.log.debug("notice window command: " + command_name)
--- a/TypeScript.py
+++ b/TypeScript.py
@@ -10,8 +10,8 @@ import sublime
 import sublime_plugin
 
 ''' Enable logging '''
-logFileLevel = logging.DEBUG
-logConsLevel = logging.DEBUG
+logFileLevel = logging.WARN
+logConsLevel = logging.WARN
 
 def set_log_level(logger):
     logger.logFile.setLevel(logFileLevel)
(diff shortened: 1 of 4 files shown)


zhengbli:
Copy my comment over:

So, after some more investigation, I believe the most possible reason why the editor hanged when batch closing happens is unnecessary file opening.

The logging shows when closing a file, the following event listeners are triggered in order:

1. ``on_activated`` (only triggered when using "close all" command)
2. ``on_close``

And both methods call ``getInfo``, which tries to open the file on the server. Therefore if any of the files to be closed has never been opened on server before, the plugin will try to do that right before closing it, which involves creating a temp project system for each file and is expensive and unnecessary. And to make things worse, although sublime appears to remember all the files in a window after you close and reopen the window, it doesn't load them until you activate the view. So it is not hard to explain why the hang happens a lot. 

So my proposal tries to solve the two common cases of batch closing:
- via ``close all`` window command:
In this case sublime will trigger ``on_window_command`` event listener. So I set a global flag in this listener to notify the following closings that this is a batch closing, so don't try to open the files;

- via clicking on the red cross button on the window's title bar:
In this case ``on_window_command`` event listener won't be triggered. However, sublime will not trigger ``on_activated`` on each view either. Therefore any unopened file will remain unopened when ``on_close`` is called. And I can just avoid unopened files in ``on_close``.

Please comment if there are anything wrong in my attempt to solve this problem, or if there are better solutions. Thanks!

TypeScript.py
```diff
@@ -8,6 +9,29 @@
 import sublime
 import sublime_plugin
 
+''' Enable logging '''
+logFileLevel = logging.DEBUG
```
billti: You might want to make the file log level lower by default (maybe logging.WARN), else if using Sublime Text for long TypeScript sessions without quitting, this might end up being quite a large file.

TypeScript.py
```diff
@@ -8,6 +9,29 @@
...
+    logger.console.setLevel(logConsLevel)
+
+# Sublime/Python 2 & 3 differ in the name of this module, thus package import
+# needs to be handled slightly differently
+if sys.version_info < (3, 0):
+    from libs import logger
+    set_log_level(logger)
+
+    # This package exposes the Sublime classes that need to be reexposed here
+    from libs import *
```
billti: You don't need to copy this bit (or the equivalent on line 31).  This is only here as in my original code I'm exporting the Sublime Commands classes from 'libs', which need to be reexported here.  You only need logger, which you imported in line 23.

TypeScript.py
```diff
@@ -469,19 +495,22 @@ def change_count(self, view):
...
-            info.prevSel = copyRegionsStatic(view.sel())
-            # ask server for initial error diagnostics
-            self.refreshErrors(view, 200)
-            # set modified and selection idle timers, so we can read
-            # diagnostics and update
-            # status line
-            self.setOnIdleTimer(20)
-            self.setOnSelectionIdleTimer(20)
-            self.changeFocus = True
+        # logger.view_debug(view, "enter on_activated")
```
billti: This, and several other logging lines below, you don't need to comment out.  You already have them at log level at 'DEBUG'.  Just change the level to 'INFO' or higher if you won't want to see them.

billti:
Left a few cleanup comments, but the logic look good.  Thanks! :+1:

PR #123, opened 2015-04-07
Fix completion start with dollar

ukyo:
String disappears when I select a completion start with dollar.

### DIFF
--- a/TypeScript.py
+++ b/TypeScript.py
@@ -846,7 +846,7 @@ class TypeScriptListener(sublime_plugin.EventListener):
             if rawCompletions:
                 for rawCompletion in rawCompletions:
                     name = rawCompletion.name
-                    completion = (name + "\t" + rawCompletion.kind, name)
+                    completion = (name + "\t" + rawCompletion.kind, name.replace("$", "\$"))
                     completions.append(completion)
                 self.pendingCompletions = completions
             else:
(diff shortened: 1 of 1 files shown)


TypeScript.py
```diff
@@ -846,7 +846,7 @@ def handleCompletionInfo(self, completionsResp):
             if rawCompletions:
                 for rawCompletion in rawCompletions:
                     name = rawCompletion.name
-                    completion = (name + "\t" + rawCompletion.kind, name)
+                    completion = (name + "\t" + rawCompletion.kind, name.replace("$", "\$"))
```
DanielRosenwasser: Use `"\\$"`.

DanielRosenwasser:
Not sure what this PR is supposed to do. Can you please clarify?

billti:
Agree with Daniel, can you describe what is currently broken?  Is this related to $ potentially being interpreted as a snippet marker and thus needing to be escaped?  Please open an issue that describes the problem, and reference the relevant issue from any pull requests.

DanielRosenwasser:
Context from [the link](http://www.sublimetext.com/forum/viewtopic.php?f=2&p=37311) given in #125:

> In snippets (rather than completions-files) dollar-signs need to be double-escaped \\ - but I should mention/apologise that I haven't proven my solution

DanielRosenwasser:
:+1: I just pulled this down and it looks good; `replace` actually doesn't allocate a new string unless `$` occurs in the original name, so at least we avoid that overhead.

Any thoughts @billti?

zhengbli:
I tried this solution with the dollar sign in different positions, and it all worked fine. :+1: 
Although if I start typing after the dollar sign, the completion list is still not right. However that should be a different issue, I'll file it elsewhere.

PR #128, opened 2015-04-08
Added signature popup support on ST3

billti:
Getting this up in early form for folks to play with.  I've only tested on ST3 on OS X so far.  I will verify nothing has broken on ST2 and other platforms tomorrow.

On ST3, I hijack the Quick Info trigger ( ctrl-t, ctrl-q ) and show a signature popup instead.  There is also convenience shortcut; by just pressing a period '.' char inside the opening paren of the arg list the popup will also trigger (open to feedback on this one).  A popup will then appear with signature help as shown below.

As you type, the arg list will update.  You can also scroll through the signatures by pressing alt+up/alt+down, and if you click on the count link (i.e. "1/6" below) it will open the quick panel with all signatures so they can be fuzzy searched quickly.

I'll do some more cleanup, validation, and integration with Steve's changes tomorrow, but this should be good enough to get a feel for now.

screen shot 2015-04-07 at 11 18 43 pm


TypeScript.py
```diff
@@ -47,6 +48,12 @@ def set_log_level(logger):
 
 TOOLTIP_SUPPORT = int(sublime.version()) >= 3072
 
+# Contains the template resource used to style signature overload popups
+popup_html = None
+
+# Contains the popup session when active
+popup_session = None
+
```
zhengbli: It might be easier to refactor later to set "popup_html" and "popup_session" as static members of your command class, instead of a global variable.

TypeScript.py
```diff
@@ -986,6 +1030,169 @@ def on_selected(self, index):
                               {"contents": self.snippets[index]})
 
 
+class TypescriptSignaturePopup(sublime_plugin.TextCommand):
+    def is_enabled(self):
+        return is_typescript(self.view)
+
+    def run(self, edit, move=None):
+        print('In run for signature panel')
```
zhengbli: as we already have a logging module, may be easier to control the output if replacing print with logger.log.debug or other level

popup.html
```diff
@@ -0,0 +1,33 @@
+<style>
+html {
+    background-color: #232628;
+    color: #CCCCCC;
```
danquirk: Is there a way to make the color scheme in here respect your existing color scheme somehow?

billti:
OK, after about the 4th refactor :weary:, I'm pretty happy with where this is right now.

With the initial implementation, requests were taking about 180ms - 200ms for signature help on heavy weight signatures such as addEventListener.  As these were synchronous and requested on every event, this was quite janky.  With this latest round, I've made the following updates:

 - Implemented a WorkScheduler class to throttle requests.  This effectively defers requests for a period (which gets longer as the requests do), and only sends the last request at the end of this period.  This got rid of most of the jank and worked quite nicely, though the requests themselves still took just as long.  This is well commented if interested.
 - Switched to using async requests.  This got the total time per request down a little to about 120 - 150ms, but as the UI thread now isn't blocking during a chunk of this, the experience improved again.
 - Switched to using straight JSON from the response instead of constructing the class hierarchy.  This got requests all the way down to 30 - 60ms consistently for addEventListener... this is HUGE!  Constructing those classes is really expensive for structures the size of SignatureHelpResponse.

So we're down from ~180ms per (frequent) blocking request, to about 40ms for infrequent non-blocking requests for addEventListener - which is a massive gain.  Most requests (which are way smaller) barely even register now.

I'm really happy with where this is perf wise now.  Just gotta bash on it and get out the bugs and ship it! :smile: 

A couple minor changes: Signature help auto appears on left-paren now.  Also, to trigger it inside an arg list use Alt+Comma, which matches nicely with Alt+Up/Down to cycle through the list.

I've been testing on ST3 on Window so far.  I'll verify others later, so your mileage may vary at this point.

zhengbli:
I was trying it out, but it seems to have some problem loading the html template, only blank tooltips are showing up (the speed is pretty good).

The screen shot:
untitled

And the log content:
```
DPI scale: 1
startup, version: 3083 windows x64 channel: stable
executable: /C/Program Files/Sublime Text 3/sublime_text.exe
working dir: /C/Program Files/Sublime Text 3
packages path: /C/Users/Zhengbo/AppData/Roaming/Sublime Text 3/Packages
state path: /C/Users/Zhengbo/AppData/Roaming/Sublime Text 3/Local
zip path: /C/Program Files/Sublime Text 3/Packages
zip path: /C/Users/Zhengbo/AppData/Roaming/Sublime Text 3/Installed Packages
ignored_packages: ["Vintage"]
pre session restore time: 0.124615
startup time: 0.140615
first paint time: 0.155615
reloading plugin Default.block
reloading plugin Default.comment
reloading plugin Default.copy_path
reloading plugin Default.delete_word
reloading plugin Default.detect_indentation
reloading plugin Default.duplicate_line
reloading plugin Default.echo
reloading plugin Default.exec
reloading plugin Default.fold
reloading plugin Default.font
reloading plugin Default.goto_line
reloading plugin Default.history_list
reloading plugin Default.indentation
reloading plugin Default.kill_ring
reloading plugin Default.mark
reloading plugin Default.new_templates
reloading plugin Default.open_context_url
reloading plugin Default.open_file_settings
reloading plugin Default.open_in_browser
reloading plugin Default.pane
reloading plugin Default.paragraph
reloading plugin Default.paste_from_history
reloading plugin Default.quick_panel
reloading plugin Default.save_on_focus_lost
reloading plugin Default.scroll
reloading plugin Default.set_unsaved_view_name
reloading plugin Default.side_bar
reloading plugin Default.sort
reloading plugin Default.swap_line
reloading plugin Default.switch_file
reloading plugin Default.symbol
reloading plugin Default.transform
reloading plugin Default.transpose
reloading plugin Default.trim_trailing_white_space
reloading plugin CSS.css_completions
reloading plugin Diff.diff
reloading plugin HTML.encode_html_entities
reloading plugin HTML.html_completions
reloading plugin 0_package_control_loader.00-package_control
reloading plugin 0_package_control_loader.02-bz2
reloading plugin Package Control.bootstrap
reloading plugin Package Control.Package Control
reloading plugin PackageResourceViewer.package_resource_viewer
reloading plugin PackageResourceViewer.package_resources
reloading plugin TypeScript.TypeScript
2015-04-11 14:16:08,405: 2972: INFO: Logging configured to log to file: C:\Users\Zhengbo\AppData\Roaming\Sublime Text 3\Packages\TypeScript\TS.log
2015-04-11 14:16:08,405: 2972: WARNING: TypeScript plugin initialized.
2015-04-11 14:16:08,421: 2972: INFO: Logging configured to log to file: C:\Users\Zhengbo\AppData\Roaming\Sublime Text 3\Packages\TypeScript\TS.log
reloading plugin User.hello_world
plugins loaded
initialize typescript...
3083
spawning node module: C:\Users\Zhengbo\AppData\Roaming\Sublime Text 3\Packages\TypeScript\tsserver\tsserver.js
Found node executable at node
2015-04-11 14:16:08,765: 2972: INFO: Popup resource path: Packages/TypeScript/popup.html
2015-04-11 14:16:08,765: 2972: INFO: Loaded tooltip template from Packages/TypeScript/popup.html
ref view not found
2015-04-11 14:16:08,765: 4524: INFO: 0 byte line in stream when expecting header
Package Control: Skipping automatic upgrade, last run at 2015-04-11 13:56:00, next run at 2015-04-11 14:56:00 or after
loaded C:\Users\Zhengbo\Desktop\test.ts
insert_snippet
time for reload 0.015625
2015-04-11 14:16:16,687: 4524: INFO: 0 byte line in stream when expecting header
2015-04-11 14:16:16,781: 4524: INFO: 0 byte line in stream when expecting header
2015-04-11 14:16:17,484: 4524: INFO: 0 byte line in stream when expecting header
```

billti:
Interesting... you can see from the log it is loading the popup file fine (the 'Loaded tooltip template...' line).  Also, the popup is actually laid out per the DIVs in the template, so it seems just the CSS isn't taking effect.  The CSS is actually in the same template however.

Do you have any other plugins or custom themes that might be interfering here?

FWIW: I just cloned the repo on my home machine and checked out the tooltips branch (didn't have this plugin set up before on this box), and it's working fine.

zhengbli:
Yeah it is kind of odd ... I reset my sublime to clean mode and synced the tooltip branch again, and it remained the same.

zhengbli:
I think the problem is in ``sublime.load_resource``. I changed that to simply ``open``, and then the CSS loaded fine.

TypeScript.py
```diff
@@ -1013,6 +1050,24 @@ def on_selected(self, index):
         self.view.run_command('insert_snippet',
                               {"contents": self.snippets[index]})
 
+class TypescriptSignaturePopup(sublime_plugin.TextCommand):
+    def is_enabled(self):
+        return is_typescript(self.view)
+
```
billti: Add TOOLTIP_SUPPORT check also.

libs/nodeclient.py
```diff
@@ -15,6 +16,8 @@
 import jsonhelpers
 import servicedefs
 
+_read_counter = 0
```
billti: Move this. From old debugging investigation.

libs/nodeclient.py
```diff
@@ -5,6 +5,7 @@
 import json
 import sublime
 import sublime_plugin
+from logger import log
 
 # queue module name changed from Python 2 to 3
 try:
```
billti: Change the import to detect Python version, rather than handle exception.

libs/nodeclient.py
```diff
@@ -175,34 +178,49 @@ def getEvent(self):
 
     @staticmethod
     def __readMsg(stream, msgq, eventq, asyncReq):
+        global _read_counter
```
billti: Delete

libs/workscheduler.py
```diff
@@ -0,0 +1,161 @@
...
+        log.debug('In cancel for work scheduler')
+        with self.lock:
+            self.canceled = True
+            self.next_job = None
+            self.last_time = 0
+            self.last_cost = 0
+            self.timer_set = False
+
+
+_default_scheduler = WorkScheduler()
```
billti: Can delete here and below.  PopupManager doesn't use the default.

popup.html
billti: I can look into that.  I think you may have to handle that logic yourself.  I'll add that as an issue to follow up on.

libs/popupmanager.py
```diff
@@ -0,0 +1,209 @@
...
+                                    arg_span["end"]["offset"] - 1)
+        arg_region = sublime.Region(span_start, span_end)
+        view.add_regions('argSpan', [arg_region],
+                         flags=sublime.HIDDEN)
+        # To view region, set to: scope='comments', flags=sublime.DRAW_EMPTY)
+
+        self.display()
+
+    def display(self):
+        popup_parts = self.get_current_signature_template()
```
billti: Maybe named better: get_signature_parts?

libs/popupmanager.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def on_close_popup(self):
+        """ Call whenever the view loses focus, of the region is exited """
+        if self.current_view:
+            self.current_view.erase_regions('argSpan')
+            self.current_view.hide_popup()
+            self.current_view = None
+        self.signature_help = None
+
+    def is_active(self):
```
billti: Comment this.  Just means a session is currently active.  current_view is set iff this is true.

libs/workscheduler.py
```diff
@@ -0,0 +1,199 @@
...
+
+            This pulls the queued work-item from self.next_job, which is an
+            instance of 'do_work' above, and executes it.
+            """
+            log.debug('In on_scheduled for work scheduler')
+            job = None
+            job_running = False
+            with self.lock:
+                if self.job_running:
+                    job_running = True
```
zhengbli: just ``if not self.job_running``

libs/workscheduler.py
```diff
@@ -0,0 +1,199 @@
...
+            end_time = time.time()
+            canceled = False
+            with self.lock:
+                self.last_cost = end_time - self.last_time
+                self.job_running = False
+                canceled = self.canceled
+            log.debug('Work took {0:d}ms'.format(int(self.last_cost * 1000)))
+            if not canceled:
+                # Post the response to the handler on the main thread
+                sublime.set_timeout(lambda: handler(results), 0)
```
zhengbli: If the timeout is 0, why not call ``handler(results)`` directly?

billti: See that comment on the line above it? :-)  That doesn't happen if you call it directly.

zhengbli:
:+1: :ship: :it:

NekR:
How to call this popup manually? It appears only when I type "(" for method/function.

billti:
It's Alt+Comma. (see https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/master/Default.sublime-keymap#L77 ).  We backed out the readme updates due to them publishing to package control before we had published the new bits.  We'll add this back.

NekR:
@billti Thanks! I completely forgot what it's possible to get it from config.

zakdances:
I don't understand how to use this. The signature popup appears, but how do you insert the snippet? Where is the autocomplete with tabbable fields?

zhengbli:
@zakdances it doesn't insert snippets. This feature only shows the function signatures, and underline the parameter that you are currently inputting.

PR #132, opened 2015-04-08
Initial implementation of Nav to

zhengbli:
The command works with a quick panel, and redo the query every time the user types something. The user can use up / down arrow key to select the item, and hit enter to navigate to the chosen location.

For now the shortcut is set as "ctrl + f7"
demo

### DIFF
--- a/TypeScript.py
+++ b/TypeScript.py
@@ -1555,7 +1555,7 @@ class NavToCommand(sublime_plugin.WindowCommand):
         NavToCommand.started = True
         actual_search_text = "a" if search_text == "" else search_text
         self.items = cli.service.navTo(actual_search_text, self.window.active_view().file_name())
-        self.window.show_quick_panel(self.format_navto_res(self.items), self.on_done, flags = sublime.KEEP_OPEN_ON_FOCUS_LOST, on_highlight = self.on_highlight)  
+        self.window.show_quick_panel(self.format_navto_res(self.items), self.on_done)  
         logger.log.debug ("end running nvato wiht text: %s" % search_text)
         
     def on_done(self, index):
--- a/TypeScript.py
+++ b/TypeScript.py
@@ -1582,7 +1582,7 @@ class NavToCommand(sublime_plugin.WindowCommand):
 
     def format_navto_res(self, item_list):
         return [
-                [i['kind'] + ": " + i['name'], 
+                [i['name'] + ": " + i['kind'], 
                 "in %s %s" % (
                     (i["containerKind"] if "containerKind" in i else os.path.basename(i["file"]) + " (global)"), 
                     (i["containerName"] if "containerName" in i else "")
--- a/TypeScript.py
+++ b/TypeScript.py
@@ -490,7 +490,14 @@ class TypeScriptListener(sublime_plugin.EventListener):
 
     # called by Sublime when a view receives focus
     def on_activated(self, view):
-        logger.view_debug(view, "enter on_activated")
+        logger.view_debug(view, "enter on_activated " + str(view.id()))
+        if view.window() and view.window().active_view().id() != view.id():
+            if NavToCommand.started:
+                # the current view is the nav to bar
+                NavToCommand.ini_finished = False
+                view.run_command("insert_text", {"text": NavToCommand.search_text})
+                NavToCommand.ini_finished = True
+
         if not self.about_to_close_all:
             info = self.getInfo(view)
             if info:
@@ -505,7 +512,7 @@ class TypeScriptListener(sublime_plugin.EventListener):
                 self.setOnIdleTimer(20)
                 self.setOnSelectionIdleTimer(20)
                 self.changeFocus = True
-            logger.view_debug(view, "exit on_activated")
+        logger.view_debug(view, "exit on_activated " + str(view.id()))
 
     # ask the server for diagnostic information on all opened ts files in
     # most-recently-used order
@@ -760,33 +767,48 @@ class TypeScriptListener(sublime_plugin.EventListener):
     # usually called by Sublime when the buffer is modified
     # not called for undo, redo
     def on_modified(self, view):
-        info = self.getInfo(view)
-        if info:
-            info.modified = True
-            if cli.ST2():
-               info.modCount+=1
-            info.lastModChangeCount = self.change_count(view)
-            self.mod = True
-            (lastCommand, args, rept) = view.command_history(0)
-#            print("modified " + view.file_name() + " command " + lastCommand + " args " + str(args) + " rept " + str(rept))
-            if info.preChangeSent:
-                # change handled in on_text_command
-                info.clientInfo.changeCount = self.change_count(view)
-                info.preChangeSent = False
-            elif lastCommand == "insert":
-                if (not "\n" in args['characters']) and info.prevSel and (len(info.prevSel) == 1) and (info.prevSel[0].empty()) and (not info.clientInfo.pendingChanges):
+        logger.log.debug ("enter on_modified " + str(view.id()))
+        # it is a special view
+        if view.window() and view.window().active_view().id() != view.id():
+            logger.log.debug ("enter on_modified: special view. started: %s, init_finished: %s" % 
+                   (NavToCommand.started, NavToCommand.ini_finished))
+
+            if NavToCommand.started and NavToCommand.ini_finished:
+                new_content = view.substr(sublime.Region(0, view.size()))
+                sublime.active_window().run_command("nav_to", {'search_text': new_content})
+
+            logger.log.debug ("exit on_modified: special view. started: %s, init_finished: %s" % 
+                   (NavToCommand.started, NavToCommand.ini_finished))
+        # it is a normal view
+        else:
+            info = self.getInfo(view)
+            if info:
+                info.modified = True
+                if cli.ST2():
+                   info.modCount+=1
+                info.lastModChangeCount = self.change_count(view)
+                self.mod = True
+                (lastCommand, args, rept) = view.command_history(0)
+    #            print("modified " + view.file_name() + " command " + lastCommand + " args " + str(args) + " rept " + str(rept))
+                if info.preChangeSent:
+                    # change handled in on_text_command
                     info.clientInfo.changeCount = self.change_count(view)
-                    prevCursor = info.prevSel[0].begin()
-                    cursor = view.sel()[0].begin()
-                    key = view.substr(sublime.Region(prevCursor, cursor))
-                    sendReplaceChangesForRegions(view, staticRegionsToRegions(info.prevSel), key)
-                    # mark change as handled so that on_post_text_command doesn't
-                    # try to handle it
-                    info.changeSent = True
-                else:
-                    # request reload because we have strange insert
-                    info.clientInfo.pendingChanges = True
-            self.setOnIdleTimer(100)
+                    info.preChangeSent = False
+                elif lastCommand == "insert":
+                    if (not "\n" in args['characters']) and info.prevSel and (len(info.prevSel) == 1) and (info.prevSel[0].empty()) and (not info.clientInfo.pendingChanges):
+                        info.clientInfo.changeCount = self.change_count(view)
+                        prevCursor = info.prevSel[0].begin()
+                        cursor = view.sel()[0].begin()
+                        key = view.substr(sublime.Region(prevCursor, cursor))
+                        sendReplaceChangesForRegions(view, staticRegionsToRegions(info.prevSel), key)
+                        # mark change as handled so that on_post_text_command doesn't
+                        # try to handle it
+                        info.changeSent = True
+                    else:
+                        # request reload because we have strange insert
+                        info.clientInfo.pendingChanges = True
+                self.setOnIdleTimer(100)
+        logger.log.debug ("exit on_modified " + str(view.id()))
 
     # ST3 only
     # called by ST3 for some, but not all, text commands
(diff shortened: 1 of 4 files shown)


steveluc:
@zhengbli this is great.  Could you please put in the container?  As in:
getQuickInfo                         method     in Session class
getQuickInfoAtPosition         method     in LanguageService interface
getQuickInfoAtPosition         function    in createLanguageService function

To disambiguate the symbols, the filename is not very helpful and can be dropped.  Instead it is very useful to have the container in addition to the kind (which you already have).  Also, could you please put the name first as in the above?  It is easier to scan through the names that way.

steveluc:
PS the file can be the container in the case of global variables, functions, types but I think the leaf filename is sufficient (you could put in one level of directory if there are two files of the same leaf name containing the identical symbol name).  That way the whole thing can go on one line and we can see more symbols.

zhengbli:
Is this what you envisioned?
changed

steveluc:
That is great!  Thanks.  :+1:

steveluc:
Will you please put the names first?  It is easier to read them down the left side of the list.
createTFoot    (method in HTMLTableElement interface)
deleteTFoot    (method in HTMLTableElement interface)
tFoot               (property   in HTMLTableElement interface)

zhengbli:
I put the names first, but still left the container information in the second line, because they can easily become quite long

steveluc:
OK.  :+1:

TypeScript.py
```diff
@@ -1501,6 +1523,81 @@ def run(self, text):
             formatRange(text, view, rblineStart, ralineEnd)
 
 
+class NavToCommand(sublime_plugin.WindowCommand):
+    # indicate the nav to panel is started or not
```
DanielRosenwasser: `# indicate whether the nav panel is started or not`

TypeScript.py
```diff
@@ -1501,6 +1523,81 @@ def run(self, text):
...
+    started = False
+    # indicate if the insert_text command has finished pasting text into the textbox.
+    # during which time the on_modified callback shouldn't do angthing
+    ini_finished = False
```
DanielRosenwasser: ini?

TypeScript.py
```diff
@@ -1501,6 +1523,81 @@ def run(self, text):
...
+    @classmethod
+    def reset(cls):
+        cls.started = False
+        cls.ini_finished = False
+
+    def is_enabled(self):
+        return is_typescript(self.window.active_view())
+
+    def run(self, search_text = ""):
+        logger.log.debug ("start running nvato wiht text: %s" % search_text)
```
DanielRosenwasser: navto with

TypeScript.py
```diff
@@ -1501,6 +1523,81 @@ def run(self, text):
             formatRange(text, view, rblineStart, ralineEnd)
 
 
+class NavToCommand(sublime_plugin.WindowCommand):
+    # indicate the nav to panel is started or not
+    started = False
+    # indicate if the insert_text command has finished pasting text into the textbox.
+    # during which time the on_modified callback shouldn't do angthing
+    ini_finished = False
+    reset_already = False
```
DanielRosenwasser: What does this indicate?

TypeScript.py
```diff
@@ -1501,6 +1523,81 @@ def run(self, text):
...
+        if NavToCommand.reset_already:
+            NavToCommand.reset_already = False
+        else:
+            logger.log.debug ("reset in run")
+            NavToCommand.reset()
+            NavToCommand.reset_already = True
+
+        NavToCommand.search_text = search_text        
+        NavToCommand.started = True
+        actual_search_text = "a" if search_text == "" else search_text
```
DanielRosenwasser: How about `query_text` and add a comment on why you need to do this.

TypeScript.py
```diff
@@ -1501,6 +1523,81 @@ def run(self, text):
...
+        self.items = cli.service.navTo(actual_search_text, self.window.active_view().file_name())
+        self.window.show_quick_panel(self.format_navto_res(self.items), self.on_done)  
+        logger.log.debug ("end running nvato wiht text: %s" % search_text)
```
DanielRosenwasser: navto with

TypeScript.py
```diff
@@ -1501,6 +1523,81 @@ def run(self, text):
...
+
+        if index >= 0:
+            item = self.items[index]
+            line, offset = item['start']['line'], item['start']['offset']
+            self.window.open_file(item['file'] + ":%s:%s" % (line, offset), sublime.ENCODED_POSITION)
+
+        logger.log.debug ("exit on_done. search_text: %s, started: %s, init_finished: %s" % 
+               (NavToCommand.search_text, NavToCommand.started, NavToCommand.ini_finished))
+
+    def format_navto_res(self, item_list):
```
DanielRosenwasser: format_navto_result

TypeScript.py
```diff
@@ -760,33 +767,48 @@ def on_selection_modified(self, view):
...
-                    # mark change as handled so that on_post_text_command doesn't
-                    # try to handle it
-                    info.changeSent = True
-                else:
-                    # request reload because we have strange insert
-                    info.clientInfo.pendingChanges = True
-            self.setOnIdleTimer(100)
+                    info.preChangeSent = False
+                elif lastCommand == "insert":
+                    if (not "\n" in args['characters']) and info.prevSel and (len(info.prevSel) == 1) and (info.prevSel[0].empty()) and (not info.clientInfo.pendingChanges):
```
DanielRosenwasser: Turn this into a variable, split it across multiple lines.

Name it something like `should_foo_blah_bleep`

TypeScript.py
```diff
@@ -760,33 +767,48 @@ def on_selection_modified(self, view):
...
+                        key = view.substr(sublime.Region(prevCursor, cursor))
+                        sendReplaceChangesForRegions(view, staticRegionsToRegions(info.prevSel), key)
+                        # mark change as handled so that on_post_text_command doesn't
+                        # try to handle it
+                        info.changeSent = True
+                    else:
+                        # request reload because we have strange insert
+                        info.clientInfo.pendingChanges = True
+                self.setOnIdleTimer(100)
+        logger.log.debug ("exit on_modified " + str(view.id()))
```
DanielRosenwasser: Remove space before left paren

TypeScript.py
```diff
@@ -1501,6 +1523,81 @@ def run(self, text):
...
+            NavToCommand.reset_already = False
+        else:
+            logger.log.debug ("reset in on_done")
+            NavToCommand.reset()
+            NavToCommand.reset_already = True
+
+        if index >= 0:
+            item = self.items[index]
+            line, offset = item['start']['line'], item['start']['offset']
+            self.window.open_file(item['file'] + ":%s:%s" % (line, offset), sublime.ENCODED_POSITION)
```
DanielRosenwasser: Factor `item['file'] + ":%s:%s" % (line, offset)` into a variable.

DanielRosenwasser: OR preferably, make a command that formats this for you and use it everywhere you call this

libs/serviceproxy.py
```diff
@@ -131,8 +132,27 @@ def getEvent(self):
...
+    def navTo(self, searchText, fileName):
+        req = self.wrap_req_dict ({
+            'command': 'navto',
+            'arguments': {
+                'searchValue': searchText,
+                'file': fileName,
+                'maxResultCount': 20
+            }
+        })
+        resp_str = self.__comm.sendCmdSync(json.dumps(req), req['seq'])
```
DanielRosenwasser: response_str

DanielRosenwasser: How is this a response *string* if you're indexing it on the next line?

libs/serviceproxy.py
```diff
@@ -131,8 +132,27 @@ def getEvent(self):
...
+        if resp_str['success']:
```
DanielRosenwasser: Why doesn't anything else in this file check for success?

libs/serviceproxy.py
```diff
@@ -131,8 +132,27 @@ def getEvent(self):
...
+            return resp_str['body']
+        else:
+            return {}
+
+    def wrap_req_dict(self, request_dict):
```
DanielRosenwasser: Is the idea that this is faster than creating an object? Why don't you just manufacture the whole object in this function?

DanielRosenwasser: Or even, why not create the object in the function?

TypeScript.py
```diff
@@ -760,33 +767,48 @@ def on_selection_modified(self, view):
...
-#            print("modified " + view.file_name() + " command " + lastCommand + " args " + str(args) + " rept " + str(rept))
-            if info.preChangeSent:
-                # change handled in on_text_command
-                info.clientInfo.changeCount = self.change_count(view)
-                info.preChangeSent = False
-            elif lastCommand == "insert":
-                if (not "\n" in args['characters']) and info.prevSel and (len(info.prevSel) == 1) and (info.prevSel[0].empty()) and (not info.clientInfo.pendingChanges):
+        logger.log.debug ("enter on_modified " + str(view.id()))
+        # it is a special view
+        if view.window() and view.window().active_view().id() != view.id():
```
DanielRosenwasser: Make a function `is_special_view` that takes `view.window()` and `view.id()` as arguments.

libs/serviceproxy.py
zhengbli: As we are moving to remove the whole object definition layer, I want to try using the dictionary pattern for nav to command. And I made the function because I figure the adding sequence id and message type part can be reused by other methods.

libs/serviceproxy.py
zhengbli: Good question. The "sendCmdSync" has a try / exception block that makes timeout message if no response is received. My guess is that if a request failed, no response would return; so it might not be necessary to explicitly check the 'success' property. I'll check with others to verify

TypeScript.py
zhengbli: I didn't touch this. What I did was wrap the original code in a if else block

TypeScript.py
```diff
@@ -1501,6 +1531,92 @@ def run(self, text):
...
 
+class NavToCommand(sublime_plugin.WindowCommand):
+    navto_panel_started = False
+
+    # indicate if the insert_text command has finished pasting text into the textbox.
+    # during which time the on_modified callback shouldn't do angthing
+    insert_text_finished = False
+
+    # when a new search string comes in, the command state needs to be reset and then
+    # restart. However the reset cannot be reset after the "run" method. Because
```
DanielRosenwasser: `restart` -> `restarted`

What is the difference between a reset and restart?

TypeScript.py
```diff
@@ -1501,6 +1531,92 @@ def run(self, text):
...
+    # the exeuation order of the "run" and "on_done" methods are not fixed, this flag is
+    # set to make sure the state reset to be done before run
```
DanielRosenwasser: Let's talk in person tomorrow so I can understand what's going on here a little better.

TypeScript.py
```diff
@@ -74,6 +74,14 @@ def is_typescript_scope(view, scopeSel):
 
     return view.match_selector(location, scopeSel)
 
+def is_special_view(cur_view):
+    """
+    Determine if the current view is a special view (quick panel to be the most 
+    common ones). active_view can only be normal views, therefore the id shouldn't
```
DanielRosenwasser: "(most commonly, the quick panel)"

DanielRosenwasser: This docstring comment should explain what a special view is. The second sentence should be a plain comment and explain why your logic works. For instance:

```Python
# 'active_view' can only be a normal view, therefore if its 'id' does not match
# that of the current view, there must be special view in focus.
```

steveluc:
Please check this in.  :+1:

TypeScript.py
```diff
@@ -760,33 +775,48 @@ def on_selection_modified(self, view):
...
+
+            logger.log.debug("exit on_modified: special view. started: %s, insert_text_finished: %s" % 
+                   (NavToCommand.navto_panel_started, NavToCommand.insert_text_finished))
+        # it is a normal view
+        else:
+            info = self.getInfo(view)
+            if info:
+                info.modified = True
+                if cli.ST2():
+                   info.modCount+=1
```
DanielRosenwasser: Spaces around `+=`

TypeScript.py
```diff
@@ -1501,6 +1531,92 @@ def run(self, text):
...
+    @classmethod
+    def reset(cls):
+        cls.navto_panel_started = False
+        cls.insert_text_finished = False
+
+    def is_enabled(self):
+        return is_typescript(self.window.active_view())
+
+    def run(self, input_text = ""):
+        logger.log.debug("start running navto wiht text: %s" % input_text)
```
DanielRosenwasser: with

TypeScript.py
```diff
@@ -1501,6 +1531,92 @@ def run(self, text):
...
+        else:
+            logger.log.debug("reset in run")
+            NavToCommand.reset()
+            NavToCommand.reset_already = True
+
+        NavToCommand.input_text = input_text        
+        NavToCommand.navto_panel_started = True
+
+        # query text is not always equal to input text. When no input_text is given
+        # alternative query text is used to ensure the panel stay active
```
DanielRosenwasser: I think this is misleading unless I'm misunderstanding the logic. I think you mean to say that we still want to show results even if there is no text

Does the navto search bar actually lose focus or disappear if there is no text?

TypeScript.py
```diff
@@ -1501,6 +1531,92 @@ def run(self, text):
...
+        query_text = "a" if input_text == "" else input_text
+        items = cli.service.navTo(query_text, self.window.active_view().file_name())
+        self.items = items if len(items) != 0 else self.items
+
+        self.window.show_quick_panel(self.format_navto_result(self.items), self.on_done)  
+        logger.log.debug("end running navto wiht text: %s" % input_text)
```
DanielRosenwasser: with

TypeScript.py
```diff
@@ -1501,6 +1531,92 @@ def run(self, text):
             formatRange(text, view, rblineStart, ralineEnd)
 
 
+class NavToCommand(sublime_plugin.WindowCommand):
+    navto_panel_started = False
+
+    # indicate if the insert_text command has finished pasting text into the textbox.
+    # during which time the on_modified callback shouldn't do angthing
```
DanielRosenwasser: shouldn't run

TypeScript.py
```diff
@@ -1501,6 +1531,92 @@ def run(self, text):
...
+            item = self.items[index]
+            line, offset = item['start']['line'], item['start']['offset']
+            file_at_location = item['file'] + ":%s:%s" % (line, offset)
+            self.window.open_file(file_at_location, sublime.ENCODED_POSITION)
+
+        logger.log.debug("exit on_done. input_text: %s, started: %s, insert_text_finished: %s" % 
+               (NavToCommand.input_text, NavToCommand.navto_panel_started, NavToCommand.insert_text_finished))
+
+    def format_navto_result(self, item_list):
+        return [
```
DanielRosenwasser:
```Python
return map(format_navto_item, item_list)
```

libs/serviceproxy.py
```diff
@@ -131,8 +132,27 @@ def getEvent(self):
...
+    def navTo(self, searchText, fileName):
+        req = self.wrap_req_dict ({
+            'command': 'navto',
+            'arguments': {
+                'searchValue': searchText,
+                'file': fileName,
+                'maxResultCount': 20
+            }
+        })
+        resp_dict = self.__comm.sendCmdSync(json.dumps(req), req['seq'])
```
DanielRosenwasser: `response_dict`

TypeScript.py
zhengbli: Yes, if an empty list is provided the panel will just disappear.

TypeScript.py
```diff
@@ -490,7 +498,14 @@ def change_count(self, view):
 
     # called by Sublime when a view receives focus
     def on_activated(self, view):
-        logger.view_debug(view, "enter on_activated")
+        logger.view_debug(view, "enter on_activated " + str(view.id()))
+        if is_special_view(view):
+            if NavToCommand.navto_panel_started:
+                # the current view is the nav to bar
+                NavToCommand.insert_text_finished = False
```
billti: Make this a little clearer so the code in easy to understand.  Maybe "# The current view is the QuickPanel.  Set insert_text_finished to false to suppress handling in on_modified".

TypeScript.py
```diff
@@ -1501,6 +1531,92 @@ def run(self, text):
...
+                    i['kind'],
+                    (i["containerKind"] if "containerKind" in i else os.path.basename(i["file"]) + " (global)"), 
+                    (i["containerName"] if "containerName" in i else "")
+                )] 
+                for i in item_list]
+
+    def on_highlight(self, index):
+        pass
+
+class InsertTextCommand(sublime_plugin.TextCommand):
```
billti: Do we need this?  Why not use one of the existing Sublime commands to insert/append text?

libs/serviceproxy.py
billti: Let's test this on a large project.  If this is sync, then this is blocking.  We need to make sure this isn't too sluggish.

billti:
Looks good! :+1:  I just merged the signature help popup work into master, so you might want to merge with this first and verify they play together well.  If good, merge it in.  Thanks!

zhengbli:
Yeah I merged with the ``tooltips`` branch and has both features working fine. Thanks!

PR #138, opened 2015-04-10
Fix issues due to sublime packaging

billti:


DanielRosenwasser:
Can you explain what issues this aims to fix and what the reasons for the issues were?

zhengbli:
The issue is by default sublime package control put the package in a packed file (*.sublime-package), which is essentially a zip file. This caused problem for us because we need to load resources and read files from the package. Therefore we decided to config sublime package control to install the package as an "unpacked" folder instead, which is done simply by adding a ".no-sublime-package" to the repo.

DanielRosenwasser:
Gotcha. If we're sure that's the issue, this looks good. :+1: :ship: :it:

PR #142, opened 2015-04-14
Renaming / Adding command entry / Bug fix for nav-to command

zhengbli:
Rename NavToCommand to TypescriptNavToCommand for consistency. 
Add command entry to use in command palette.

Fix the bug that navto command failed to reset the states when following
the following steps:
1. switching from the command palette to nav to quick panel
2. after the panel shows up, quit immediately
3. reopen the command palette, and start typing, a new nav to command
panel will unexpectedly show up to replace the command palette

Also, I find that in ``on_modified`` if I use a set_timeout to separate
the processing of dismissing old panel and starting new panel, all the
callbacks are invoked in order, even if the timeout is 0. (In contrast,
if I run command ``typescript_nav_to`` directly after running
``hide_overlay``, the ``on_done`` callback of the old panel will be called
AFTER the new panel shows up) therefore part of the workaround I did
earlier can be removed for simplicity.

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -24,7 +24,7 @@
     },
     {
         "keys": [ "ctrl+alt+r"],
-        "command": "nav_to",
+        "command": "typescript_nav_to",
         "context": [
             { "key": "selector", "operator": "equal", "operand": "source.ts" }
         ]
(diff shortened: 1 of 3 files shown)


steveluc:
:+1:

PR #145, opened 2015-04-14
Show error status of active view

steveluc:
Place a dot in the upper left corner (green for no errors, red for at least one error in the file).  TODO: make this work with all themes.


steveluc:
Update: changed icons to retina-sized triangles to avoid future confusion with breakpoints.  ST2 is still a dot.

TypeScript.py
```diff
@@ -352,6 +352,8 @@ def __init__(self, filename, cc):
         self.changeCountErrReq = -1
         self.lastModChangeCount = cc
         self.modCount = 0
+        self.prevVisStart = -1
+        self.prevHasErrors = False
```
zhengbli: What are these flags keeping track of?

TypeScript.py
```diff
@@ -792,6 +795,29 @@ def on_close(self, view):
                     cli.service.close(view.file_name())
         logger.view_debug(view, "exit on_close")
 
+    def handlePulse(self):
```
zhengbli: Can you elaborate the concept of "Pulse", and what does this method do exactly?

TypeScript.py
```diff
@@ -792,6 +795,29 @@ def on_close(self, view):
...
+        view = active_view()
+        info = self.getInfo(view)
+        if info:
+            self.showTopLineStatus(info, view)
+        sublime.set_timeout(self.handlePulse, 500)
+
+    def showTopLineStatus(self, info, view):
```
zhengbli: by "Top line", do you mean the title of each tab? "Top line" sounds like the first line of the content. It may help if you can put new definitions in the doc string of the method.

TypeScript.py
steveluc: The last measured visible region start position and hasErrors flag for the view.

TypeScript.py
steveluc: There is no way to detect scrolling in the Sublime API, so we need a periodic timer to check whether the view has been scrolled and to update placement of the error status icon.

TypeScript.py
steveluc: Top line means the first visible line of the buffer in the view.

TypeScript.py
```diff
@@ -792,6 +795,29 @@ def on_close(self, view):
...
+        if not self.pulse_started:
+            self.pulse_started = True
+            sublime.set_timeout(self.handlePulse, 500)
+        visStart = view.line(view.visible_region().begin())
+        if (info.prevVisStart != visStart) or (info.prevHasErrors != info.hasErrors):
+            info.prevVisStart = visStart
+            info.prevHasErrors = info.hasErrors
+            scope = "entity.name.function"
+            if info.hasErrors:
+                scope = "keyword"
```
zhengbli: We might need dedicated scope for the icons

TypeScript.py
```diff
@@ -814,6 +840,7 @@ def on_selection_modified(self, view):
                 self.setOnSelectionIdleTimer(1250)
             else:
                 self.setOnSelectionIdleTimer(50)
+            self.showTopLineStatus(info, view)
```
zhengbli: Because changes of cursor positions and selections are not likely to introduce errors, it might be better to put the method call in ``on_modified`` instead of ``on_selection_modified``

steveluc: Agree.  I just removed it entirely from on_selection_modified because the pulse timer handles it anyway.

TypeScript.py
steveluc: Yes.  Do you know how to do that?  I am wondering if we can augment whatever is the current theme with an additional theme so that we can define colors.  The colors aren't actually used in this case because the icon supplies its own colors.

steveluc: Except in ST2.

TypeScript.py
```diff
@@ -792,6 +795,29 @@ def on_close(self, view):
                     cli.service.close(view.file_name())
         logger.view_debug(view, "exit on_close")
 
+    def handlePulse(self):
+        view = active_view()
+        info = self.getInfo(view)
+        if info:
+            self.showTopLineStatus(info, view)
+        sublime.set_timeout(self.handlePulse, 500)
```
zhengbli: Why it needs to re-check every 500ms? My understanding is icon changes can only happen if the content has been modified. Also if the use is idling, we should be able to detect that and end the loop?

TypeScript.py
```diff
@@ -792,6 +795,29 @@ def on_close(self, view):
...
+
+    def showTopLineStatus(self, info, view):
+        if not self.pulse_started:
+            self.pulse_started = True
```
zhengbli: should the ``pulse_started`` be switched to ``False`` when idle?

TypeScript.py
steveluc: Problem is there is no way to detect scrolling.  Not even a text command or selection modified event.  Would love to just move the icon upon scrolling, with no timer loop.  I am open to other approaches that would not involve a gutter icon.   Don't particularly like the timer loop.

steveluc:
This is working correctly now, but does require the pulse timer.  We could instead put an error indication in the status line (N errors).  The problem is the status line is crowded with the quick info.  We could continue to put everything in the status line for ST2, but for ST3 version > 3070, we could use the popup for quick info and error text.  That would leave only the line/column and the error status on the status line.

PR #146, opened 2015-04-15
Show error status

steveluc:
This branch is renamed from showRelatedFiles.


steveluc:
This is a renamed branch.  See discussion for #145

zhengbli:
Decided to use alternative solution.

Issue #126, opened 2015-04-08
add support ENV for "node_path" setting


tgfjt:
Hi.

`"node_path": "/Users/tgfjt/bin/node"` works well.
But `"node_path": "$HOME/bin/node"` does not works.

If we can do it. It'll be useful.

DanielRosenwasser:
Sorry, could you elaborate a bit more? Are you asking us to expand environment variables in your `node_path`?

tgfjt:
Sorry! it's not enough words.

Yes, I want that expanding ENV variables in `node_path`, because I use a `Preferences.sublime-settings` between some machines.

PR #148, opened 2015-04-16
Add support to expand environment variable in node path setting

zhengbli:
related #126

### DIFF
--- a/libs/nodeclient.py
+++ b/libs/nodeclient.py
@@ -42,9 +42,21 @@ class NodeCommClient(CommClient):
         self.__eventq = queue.Queue()
 
         # start node process
-           
         pref_settings = sublime.load_settings('Preferences.sublime-settings')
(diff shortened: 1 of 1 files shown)


DanielRosenwasser:
Does this change only work for `$HOME` as a path variable?

Does this work for `%WINDOWS_STYLE_PATHS%`?

billti:
This is not addressing the customer request exactly, and is not a very flexible design.  Rather than special casing '$HOME', why not just expand any environment variables in the string provided using the API for this that works across platforms, i.e. https://docs.python.org/2/library/os.path.html#os.path.expandvars

zhengbli:
Replaced my own implementation with the python built-in ``os.path.expandvars``, thanks for @billti 's suggestion. Now it supports both mac-style "$VAR" and windows style "%VAR%".

mhegazy:
:+1:

PR #152, opened 2015-04-16
Adjusted popup positioning while editing args

billti:
This anchors the popup to the opening paren while editing args on the one line.

For multi-line arg lists, this is a little more tricky.  I'd love to be able to anchor under the current line and beneath the opening paren also, however if there is no char at that position, you can't set the location to it. The behavior I settled on was to anchor the popup under the first non-whitespace char on the line (or EOL if no such char).  This works pretty well for most scenarios.

Note that there seems to be a bug where if you are on a new line (no args yet) right before the closing paren, the call to getSignaturesForPosition is returning no results, causing the popup to disappear.  This appears to be a timing issue (sometimes it works), and unrelated to this popup tracking/positioning.  This will need to be investigated separately.

Give it a pull and let me know what you think.

### DIFF
--- a/libs/popupmanager.py
+++ b/libs/popupmanager.py
@@ -26,6 +26,9 @@ class PopupManager():
         self.signature_index = 0
         self.current_parameter = 0
 
+        # Used to avoid looping
+        self.suppress_onhidden = 0
+
     def queue_signature_popup(self, view):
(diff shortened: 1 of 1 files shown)


billti:
Will submit a new pull request with the refactoring work included.

PR #161, opened 2015-04-27
Async buffer reload

steveluc:
Make buffer reload asynchronous so that typing is never delayed by a buffer reload.

### DIFF
--- a/TypeScript.py
+++ b/TypeScript.py
@@ -458,21 +458,23 @@ def sendReplaceChangesForRegions(view, regions, insertString):
         endLocation = getLocationFromPosition(view, region.end())
         cli.service.change(view.file_name(), location, endLocation, insertString)
 
-def recvReloadAck(reloadResp):
+def recv_reload_response(reloadResp):
     if reloadResp.request_seq in cli.tempFileMap:
         tmpfile = cli.tempFileMap.pop(reloadResp.request_seq)
         if tmpfile:
             cli.tempFileList.append(tmpfile)
 
 def getTempFileName():
+    """ Get the first unused temp file name to avoid conflicts
+    """
     seq = cli.service.seq
     if len(cli.tempFileList) > 0:
-        tmpfile = cli.tempFileList.pop()
+        temp_file_name = cli.tempFileList.pop()
     else:
-        tmpfile = os.path.join(pluginDir, ".tmpbuf"+str(cli.tmpseq))
+        temp_file_name = os.path.join(pluginDir, ".tmpbuf"+str(cli.tmpseq))
         cli.tmpseq += 1
-    cli.tempFileMap[seq] = tmpfile
-    return tmpfile
+    cli.tempFileMap[seq] = temp_file_name
+    return temp_file_name
 
 # write the buffer of view to a temporary file and have the server reload it
 def reloadBuffer(view, clientInfo=None):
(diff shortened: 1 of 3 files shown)


TypeScript.py
```diff
@@ -455,21 +458,31 @@ def sendReplaceChangesForRegions(view, regions, insertString):
         endLocation = getLocationFromPosition(view, region.end())
         cli.service.change(view.file_name(), location, endLocation, insertString)
 
+def recvReloadAck(reloadResp):
```
zhengbli: What does "Ack" mean? If you mean "async", "recv_reload_async" maybe better. For new functions / methods, we should try use [python convention](https://www.python.org/dev/peps/pep-0008/#naming-conventions) (lower_case_with_underscores.)

TypeScript.py
```diff
@@ -455,21 +458,31 @@ def sendReplaceChangesForRegions(view, regions, insertString):
...
+    if reloadResp.request_seq in cli.tempFileMap:
+        tmpfile = cli.tempFileMap.pop(reloadResp.request_seq)
```
zhengbli:
```python
tmpfile = cli.tempFileMap.pop(reloadResp.request_seq, None)
```

TypeScript.py
```diff
@@ -455,21 +458,31 @@ def sendReplaceChangesForRegions(view, regions, insertString):
...
+        if tmpfile:
+            cli.tempFileList.append(tmpfile)
+
 def getTempFileName():
-   return os.path.join(pluginDir, ".tmpbuf")
+    seq = cli.service.seq
```
zhengbli: Add comment:
```python
def getTempFileName():
    """ Get the first unused temp file name to avoid conflicts
    """

```

TypeScript.py
```diff
@@ -455,21 +458,31 @@ def sendReplaceChangesForRegions(view, regions, insertString):
...
+    if len(cli.tempFileList) > 0:
+        tmpfile = cli.tempFileList.pop()
+    else:
+        tmpfile = os.path.join(pluginDir, ".tmpbuf"+str(cli.tmpseq))
+        cli.tmpseq += 1
+    cli.tempFileMap[seq] = tmpfile
```
zhengbli: Do we delete the temp files after exiting sublime? If not, because the ``tmpseq`` starts from 1 every time sublime launches, is it possible that old temp files are reused? Will that pollute anything?

TypeScript.py
```diff
@@ -455,21 +458,31 @@ def sendReplaceChangesForRegions(view, regions, insertString):
...
+    if reloadResp.request_seq in cli.tempFileMap:
+        tmpfile = cli.tempFileMap.pop(reloadResp.request_seq)
+        if tmpfile:
+            cli.tempFileList.append(tmpfile)
+
 def getTempFileName():
-   return os.path.join(pluginDir, ".tmpbuf")
+    seq = cli.service.seq
+    if len(cli.tempFileList) > 0:
+        tmpfile = cli.tempFileList.pop()
```
zhengbli: rename to ``temp_file_name``

TypeScript.py
```diff
@@ -281,6 +281,9 @@ def __init__(self):
         self.fileMap = {}
         self.refInfo = None
         self.versionST2 = False
+        self.tempFileMap = {}
+        self.tempFileList = []
```
zhengbli: What are the differences of ``tempFileMap`` and ``tmepFileList``?

steveluc: tempFileMap is a dictionary mapping request sequence number to the temporary file name used in the request.  tempFileList is a cache of available temporary file names

zhengbli: Is the mapping between seq number and temp file name used somewhere? Or is it more like a collection of "temp files used but not available now"?

steveluc: The mapping is used in the receiver function, so that upon completion of the reload, the temporary file name corresponding to the request sequence number can be re-used.

zhengbli: Cool, that makes much sense. To make it more clear, could you rename the ``tempFileList`` to something like ``available_tempfile_list``?

zhengbli:
:+1:

mhegazy:
:+1:

PR #171, opened 2015-05-01
Fix completion issue 170 plus one more

steveluc:
Make ST2 completions once again synchronous.  Add hide completion if cursor moves to before the last location that initiated completion.


zhengbli:
Tested on both S2 and S3 on my machine and it looks fine. :+1:

mhegazy:
:+1:

PR #173, opened 2015-05-01
Fix syntax highlight issues for object types

### LINKED ISSUE #162: Syntax highlighting incorrect for type literals
```typescript
protected _normalizeDatasets<A, B>(fromX: boolean): {a: A; b: B}[] {
  return Stacked.prototype._normalizeDatasets.call(this, fromX);
}
```

The typed object literal for the return typing "{a: A; b: B}" wrecks syntax highlighting for the rest of the page. I have tried putting a semicolon after the B as well and the syntax highlighting is still incorrect. Is this an issue with my syntax, or is there a bug with the plugin?

image

zhengbli:
Fix #162

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -21,7 +21,7 @@ repository:
 
   control-statement:
     name: keyword.control.ts
-    match: \b(break|catch|continue|declare|do|else|finally|for|if|return|switch|throw|try|while)\b
+    match: \b(break|catch|continue|declare|do|else|finally|for|if|return|switch|throw|try|while|with|super)\b
 
   switch-case: 
     name: case.expr.ts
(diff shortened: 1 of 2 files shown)


mhegazy:
@steveluc and @DanielRosenwasser  can you take a look at this PR.

TypeScript.YAML-tmLanguage
```diff
@@ -390,7 +394,7 @@ repository:
 
   expression-operator:
     name: keyword.operator.ts
-    match: =>|\b(delete|export|import|in|instanceof|module|new|typeof|void)\b
+    match: =>|\b(delete|export|import|in|of|instanceof|module|new|typeof|void)\b
```
DanielRosenwasser: `of` is a contextual keyword that only lights up in a `for..of` statement.

Issue #159, opened 2015-04-24
Bad initial indentation config


lucasmciruzzi:
Every time I open a .ts file, the indendation is set to spaces instead of keeping my config of tabs. In my settings I have...

    "tab_size": 4,
    "translate_tabs_to_spaces": false,
    "typescript_auto_format": false

... and in my .editorconfig I have ...

    [*]
    charset = utf-8
    end_of_line = lf
    indent_style = tab
    insert_final_newline = true
    trim_trailing_whitespace = true

So the problem comes from the TypeScript plugin.

How can I avoid this overwrite behavior the plugin has over my indentation settings?

zhengbli:
This is a known issue, similar to #147. This happens because we need to constantly sync the buffer content with the content on server, and sublime doesn't report changes when buffer content changes when it applies its algorithm to backspace. We are working on a workaround.

buu700:
+1, this is a blocker.

I wanted to switch from Arctic TypeScript for some of the nicer features in this plugin, but that isn't really an option when it mangles my code.

zhengbli:
Recent changes should fix the indentation issues. @buu700 @lucasmciruzzi could you give it a try?

buu700:
Seems to work for me. Thanks a lot!

lucasmciruzzi:
Seems to be working for me too. Thanks!

PR #179, opened 2015-05-04
Add option "typescript_auto_format" to turn off auto format + reconfiguring host format settings when they are changed

zhengbli:


steveluc:
Looks good.  We have typescript_auto_format on the key bindings; do we also need it on the class (I think it won't do any harm either way)?
    {
        "keys": [ ";" ],
        "command": "typescript_format_on_key",
        "args": { "key": ";" },
        "context": [
            { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
            { "key": "setting.typescript_auto_format", "operator": "equal", "operand": true },
            { "key": "selector", "operator": "equal", "operand": "source.ts" }
        ]
    },

zhengbli:
The reason I added this is for now there is no way to disable the auto format if the user doesn't want that (related issue #159)

zhengbli:
Oh now I see what you mean. This is redundant if we already have that in key configs. Closing.

zhengbli:
Reopen because it also contains issue fixes.

PR #182, opened 2015-05-05
Sync tsserver.js and lib.d.ts version with TypeScript repo

zhengbli:
To incorporate several bug fixes after 1.5 beta was published in the TypeScript repo. Also last time a locally built version of tsserver.js was committed. Changed them back to be the same with the ones in the TypeScript repo.


mhegazy:
This version of typescript has not been released yet. I would stick with TypeScript 1.5 beta for now.

steveluc:
I think fixes may have gone in since 1.5 beta.

mhegazy:
We have not shipped a sublime plug in release since. that means if we ship today, users will get an inconsistent behavior specially with respect to the library.. things that work fine in the editor will break at compile time, and navigating to the error locations in the editor will show different definitions based on the new library. 

We either stick with 1.5-beta that we shipped, or ship another version, say 1.5-beta2 to match it. i am fine either ways..

zhengbli:
@mhegazy  Because there are sublime-related changes (for example https://github.com/Microsoft/TypeScript/pull/3016 that fixes the indentation configuration problems) after 1.5 beta release, I think at least that bit should be incorporated. Do you think it is better for consistence if I only change that bit, and keep the rest of the content the same with 1.5 beta?

mhegazy:
The library is the one that is a bit tricky.. specially that we had 4-5 fixes since 1.5-beta. so i am fine with this plan as well.

PR #185, opened 2015-05-06
Fix syntax highlight issues for arrow functions and function parameters

zhengbli:
Fix the issue that arrow function is not highlighted correctly (credit to @billti )
a

Also moved the equal sign ``=`` to the correct scope.


TypeScript.YAML-tmLanguage
```diff
@@ -402,11 +405,11 @@ repository:
 
   relational-operator:
     name: keyword.operator.comparison.ts
-    match: ===|==|=|!=|!==|<=|>=|<>|<|>
+    match: ===|==|!=|!==|<=|>=|<>|<|>
```
billti: "!==" should go before "!=", otherwise "!=" will always match first.

TypeScript.YAML-tmLanguage
```diff
@@ -273,8 +273,10 @@ repository:
     end: '\)'
     patterns:
     - include: '#comment'
+    - include: '#parameter-name'
```
billti: This looks like a non-trivial change that could have quite an impact.  Let's chat about this in person tomorrow before you merge anything.

zhengbli: Yeah sure. I have also written a test file the listed all the common cases for scope detection which I can show you tomorrow.

TypeScript.YAML-tmLanguage
billti: Bump on this... did this ever get corrected? (Or a test added and me proved wrong? :smile: )

PR #186, opened 2015-05-07
recommend shadow clone for end users

aminroosta:
cloning the repository takes a while.
for end users a shadow clone works just fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ cd "%APPDATA%\Sublime Text 3\Packages"
 git clone --depth 1 https://github.com/Microsoft/TypeScript-Sublime-Plugin.git TypeScript
 ```
 
-(`--depth 1` downloads only the current version to reduce the clone size.)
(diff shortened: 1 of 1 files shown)


zhengbli:
So this practically makes the end user clone the repo without the history? Interesting, and I agree for users that is normally enough. @steveluc @billti what do you think?

README.md
```diff
@@ -15,14 +15,14 @@ look something like:
 
 ```
 cd ~/"Library/Application Support/Sublime Text 3/Packages"
-git clone https://github.com/Microsoft/TypeScript-Sublime-Plugin.git TypeScript
+git clone --depth=1 https://github.com/Microsoft/TypeScript-Sublime-Plugin.git TypeScript
```
zhengbli: Add a note to explain what ``--depth=1`` does. Something like this:
(``--depth=1`` downloads only the current version to reduce the clone size)

mhegazy:
:+1:

zhengbli:
Thanks for contributing!

PR #189, opened 2015-05-10
added a rule to include html syntax definition if the user writes tem…

aminroosta:
added a rule to include html syntax definition when the user writes template: ```.( in many libraries including knockoutjs and angular2 users supply the html(view) part of a `component` via a property named _template_. I have added a new rule that only matches when the user tries to write (template:``<some>html</some>``). it includes the html syntax within `` ``. the cool part about it is that other plugins like **emmet** work within `` `` scope.

here is a screen shot for knockout and angular2 :

typescript_syntax_definition-changed

It does not make a difference for those how are not using these libraries .


DavideViolante:
Please add this, it is strongly needed with Angular 2.

DanielRosenwasser:
This is pretty cool and I like the idea. I don't know how many libraries otherwise use `template` as a property name, but @billti @mhegazy what do you think?

DanielRosenwasser:
Also, I dunno why @msftclas didn't post about the CLA thing.

aminroosta:
I no longer use sublime, but if anyone is interested to add this feature my SO question might help.
Note that we can have Css syntax highlighting as well:

http://stackoverflow.com/questions/36700442/atom-html-syntax-highlight-in-template-literals-for-angular2/36817059#36817059

mhegazy:
> This is pretty cool and I like the idea. I don't know how many libraries otherwise use template as a property name, but @billti @mhegazy what do you think?

I do not see why not. my knowlege of TM language is fairly limited, so can not comment on the contents of the PR. 
This should move to https://github.com/Microsoft/TypeScript-TmLanguage though.

billti:
I already did exactly this for my Angular2 plugin prototype (my grammar applied to template strings that were the value of object properties named "template", or preceded with a "/** @html */" comment). I can point to the code tomorrow and see what you think.

Get Outlook for iOS<https://aka.ms/o0ukef>




On Mon, May 16, 2016 at 10:58 AM -0700, "Mohamed Hegazy" <notifications@github.com<mailto:notifications@github.com>> wrote:


This is pretty cool and I like the idea. I don't know how many libraries otherwise use template as a property name, but @billti<https://github.com/billti> @mhegazy<https://github.com/mhegazy> what do you think?

I do not see why not. my knowlege of TM language is fairly limited, so can not comment on the contents of the PR.
This should move to https://github.com/Microsoft/TypeScript-TmLanguage though.

billti:
FYI: Here's my comment on the Angular2 plugin issue pointing to the need to do this: https://github.com/angular/angular/issues/7482#issuecomment-196016574 . It links to my fork of the tmLanguage where I had an initial implementation (https://github.com/billti/TypeScript-TmLanguage/commit/7e62410b49df49e37d53e1759c5757547c5f9258).

Note this can get a little involved to do properly. My prototype (and many of those linked/referenced here) doesn't handle things like the string being either a template string, double quoted, or single quoted (which then needs to handle all the attribute quotes within it in different formats also), as well as any template string being broken into parts by intermediate `${expr}` parts. These would take some work.

I agree it's probably a good idea though if someone has the bandwidth to tackle it. (Note with Sublime 3 you can also nest grammars, but this uses a new custom format, and wouldn't work for Sublime 2, VSCode, TextMate, etc...).

PR #197, opened 2015-05-14
Refactor

zhengbli:
Refactor the overall structure of the codebase.


commands/error_info.py
```diff
@@ -0,0 +1,24 @@
...
+class TypescriptErrorInfo(sublime_plugin.TextCommand):
+    """
+    Command called from event handlers to show error text in status line
+    (or to erase error text from status line if no error text for location)
+    """
+    def run(self, text):
+        client_info = cli.get_or_add_file(self.view.file_name())
+        pt = self.view.sel()[0].begin()
+        error_text = ""
+        for (region, text) in client_info.errors['syntacticDiag']:
```
RyanCavanaugh: TODO: Make these loops terminate early?

commands/format.py
```diff
@@ -0,0 +1,110 @@
...
+from ..libs.texthelpers import *
+from ..libs.reference import *
+
+
+class TypescriptFormatOnKey(sublime_plugin.TextCommand):
+    """
+    Format on ";", "}", or "\n"; called by typing these keys in a ts file
+    in the case of "\n", this is only called when no completion dialogue visible
+    """
+    def is_enabled(self):
```
RyanCavanaugh: Can we move this into a common place? Seems to be repeated a lot

commands/format.py
```diff
@@ -0,0 +1,110 @@
...
+        check_update_view(self.view)
+        sel = self.view.sel()
+        if len(sel) == 1:
+            original_pos = sel[0].begin()
+            bracket_char = self.view.substr(original_pos)
+            if bracket_char != "}":
+                self.view.run_command('move_to', {"to": "brackets"})
+                bracket_pos = self.view.sel()[0].begin()
+                bracket_char = self.view.substr(bracket_pos)
+            if bracket_char == "}":
```
RyanCavanaugh: `else`

commands/signature.py
```diff
@@ -0,0 +1,82 @@
...
+from ..libs.viewhelpers import *
+from ..libs.reference import *
+
+
+class TypescriptSignaturePanel(sublime_plugin.TextCommand):
+    def is_enabled(self):
+        return is_typescript(self.view)
+
+    def run(self, text):
+        print('TypeScript signature panel triggered')
```
RyanCavanaugh: When do we use `logger.log.debug` vs `print` ?

commands/format.py
zhengbli: For each command we need to define an is_enabled function, but I can see it is indeed clearer if we put that in a base class and then inherit from it. I'll try that.

commands/signature.py
zhengbli: It is a little ambiguous right now, my understanding is for messages that we definitely will output we can use either ``logger.log.info`` or ``logger.log.warn`` or ``print``; for messages only used for debugging purpose we should use ``logger.log.debug``, that way we can turn off those messages by setting the global log level higher than ``DEBUG`` which still keep the code.

libs/globalvars.py
```diff
@@ -0,0 +1,31 @@
...
+
+# set logging levels
+LOG_FILE_LEVEL = logging.WARN
+LOG_CONSOLE_LEVEL = logging.WARN
+
+NON_BLANK_LINE_PATTERN = re.compile("[\S]+")
+VALID_COMPLETION_ID_PATTERN = re.compile("[a-zA-Z_$\.][\w$\.]*\Z")
+
+# idle time length in millisecond
+IDLE_TIME_LENGTH = 20
```
RyanCavanaugh: Should this be 200? 20 is very short

libs/reference.py
```diff
@@ -0,0 +1,141 @@
...
+    ref_info = RefInfo(first_line, ref_id)
+    ref_info.set_ref_line(current_line)
+    ref_info.set_last_line(last_line)
+    for key in dict.keys():
+        ref_info.add_mapping(key, build_ref(dict[key]))
+    return ref_info
+
+
+def highlight_ids(view, ref_id):
+    """Highlight all occurrences of refId in view"""
```
RyanCavanaugh: `refId` -> `ref_id`

commands/format.py
```diff
@@ -0,0 +1,110 @@
...
+
+from ..libs.viewhelpers import *
+from ..libs.texthelpers import *
+from ..libs.reference import *
+
+
+class TypescriptFormatOnKey(sublime_plugin.TextCommand):
+    """
+    Format on ";", "}", or "\n"; called by typing these keys in a ts file
+    in the case of "\n", this is only called when no completion dialogue visible
```
DanielRosenwasser: nit: "dialogue visible" -> "dialog is visible"

commands/format.py
```diff
@@ -0,0 +1,110 @@
...
+        return is_typescript(self.view)
+
+    def run(self, text):
+        line_region = self.view.line(self.view.sel()[0])
+        line_text = self.view.substr(line_region)
+        if NON_BLANK_LINE_PATTERN.search(line_text):
+            format_range(text, self.view, line_region.begin(), line_region.end())
+        else:
+            position = self.view.sel()[0].begin()
+            cursor = self.view.rowcol(position)
```
DanielRosenwasser: Can you change this to `line, _ = self.view.rowcol(position)`?

commands/format.py
```diff
@@ -0,0 +1,110 @@
...
+        check_update_view(view)
+        regions_before_paste = regions_to_static_regions(view.sel())
+        if IS_ST2:
+            view.add_regions("apresPaste", copy_regions(view.sel()), "", "", sublime.HIDDEN)
+        else:
+            view.add_regions("apresPaste", copy_regions(view.sel()), flags=sublime.HIDDEN)
+        view.run_command("paste")
+        regions_after_paste = view.get_regions("apresPaste")
+        view.erase_regions("apresPaste")
+        for i in range(len(regions_before_paste)):
```
DanielRosenwasser: Consider using [`zip`](https://docs.python.org/3.5/library/functions.html#zip):

```Python
for rb, ra in zip(regions_before_paste, regions_after_paste):
    # ...
```

commands/format.py
```diff
@@ -0,0 +1,110 @@
...
+            rb = regions_before_paste[i]
+            ra = regions_after_paste[i]
+            rb_line_start = view.line(rb.begin()).begin()
+            ra_line_end = view.line(ra.begin()).end()
```
DanielRosenwasser: No need to prefix these variables with `rb_` and `ra_` if there's no `rb_line_end` and `ra_line_start`.

commands/go_to_definition.py
```diff
@@ -0,0 +1,23 @@
...
+class TypescriptGoToDefinitionCommand(sublime_plugin.TextCommand):
+    """Go to definition command"""
+    def is_enabled(self):
+        return is_typescript(self.view)
+
+    def run(self, text):
+        check_update_view(self.view)
+        definitionResp = cli.service.definition(self.view.file_name(), get_location_from_view(self.view))
+        if definitionResp["success"]:
+            codeSpan = definitionResp["body"][0] if len(definitionResp["body"]) > 0 else None
```
DanielRosenwasser: There's no code path other than the `if codeSpan:`, so instead, of the conditional expression, just do:

```Python
```

DanielRosenwasser: Also, call it `definition_location` or `definitionLoc` instead of `codeSpan`. Just keep in mind that we're trying to move to snake_case_naming  as far as I understand.

commands/go_to_definition.py
```diff
@@ -0,0 +1,23 @@
...
+            if codeSpan:
+                filename = codeSpan["file"]
+                startlc = codeSpan["start"]
```
DanielRosenwasser: It's weird, to know what `lc` stands for, I have to look below.

commands/go_to_type.py
```diff
@@ -0,0 +1,25 @@
...
+
+class TypescriptGoToTypeCommand(sublime_plugin.TextCommand):
+    """Go to type command"""
+    def is_enabled(self):
+        return is_typescript(self.view)
+
+    def run(self, text):
+        check_update_view(self.view)
+        typeResp = cli.service.type(self.view.file_name(), get_location_from_view(self.view))
+        print (typeResp)
```
DanielRosenwasser: To be removed?

commands/go_to_type.py
```diff
@@ -0,0 +1,25 @@
...
+        if typeResp["success"]:
+            items = typeResp["body"]
+            if len(items) > 0:
+                codeSpan = items[0]
+                filename = codeSpan["file"]
+                startlc = codeSpan["start"]
+                sublime.active_window().open_file(
+                    '{0}:{1}:{2}'.format(filename, startlc["line"] or 0, startlc["offset"] or 0),
```
DanielRosenwasser: Wait, for go-to-type, this can line and offset can be emitted, but not for go to definition? Either this is a very bad inconsistency on the protocol or there's been a small oversight here.

commands/nav_to.py
```diff
@@ -0,0 +1,65 @@
...
+
+from ..libs import *
+from ..libs.viewhelpers import *
+from ..libs.reference import *
+
+
+class TypescriptNavToCommand(sublime_plugin.WindowCommand):
+    navto_panel_started = False
+
+    # indicate if the insert_text command has finished pasting text into the textbox.
```
DanielRosenwasser: nit: "if" -> "whether"

DanielRosenwasser: Also change the period at the end of the line to a comma

commands/nav_to.py
```diff
@@ -0,0 +1,65 @@
...
+        TypescriptNavToCommand.reset()
+
+        if index >= 0:
+            item = self.items[index]
+            line, offset = item['start']['line'], item['start']['offset']
+            file_at_location = item['file'] + ":%s:%s" % (line, offset)
+            self.window.open_file(file_at_location, sublime.ENCODED_POSITION)
+
+    def format_navto_result(self, item_list):
+        def get_description_str(i):
```
DanielRosenwasser: Can we use a name other than `i`?

commands/nav_to.py
```diff
@@ -0,0 +1,65 @@
...
+            name = i["name"]
+            kind = i["kind"]
+            container_kind = i["containerKind"] if "containerKind" in i else os.path.basename(i["file"]) + " (global)"
+            container_name = i["containerName"] if "containerName" in i else ""
+            description_str = "{0} in {1} {2}".format(kind, container_kind, container_name)
+            return [name, description_str]
+
+        return [get_description_str(i) for i in item_list]
```
DanielRosenwasser: Not sure if I ain't as Pythonic as I once was, but this *might* be clearer as a `map``.

commands/nav_to.py
DanielRosenwasser: Also, since you're not capturing anything, maybe move this out to the class.

commands/go_to_definition.py
zhengbli: but in the case when ``definitionResp["body"]`` won't it throw exception?

commands/nav_to.py
zhengbli: Thing is ``map`` doesn't by default return a list in python 3, ``list(map(get_description_str, item_list))`` seems to have similar amount of neatness to me

DanielRosenwasser: Ah, right - does this necessarily need to return a list? Where does it eventually get called from?

zhengbli: Yes it needs to return a list that feeds the quick panel for "nav to". It is called at line 37 in this file, within the ``run`` method of ``TypescriptNavToCommand``

libs/globalvars.py
zhengbli: @steveluc has a set of different idle time for different idle scenarios that I don't quite understand yet. I do think it would be clearer if we can put them all in the ``globalvars.py`` and name them properly.

zhengbli:
There is one catch that the ``line, offset`` on the server are 1-based, while the ``line, column`` used by sublime are 0-based. It might be clearer to refactor the names to ``one_based_line`` and ``zero_based_line``.

billti:
We seem to be inconsistent on the file naming.  Some have capitals in them (e.g. "TypeScript.py"), some have underscores (e.g. "go_to_type.py" or "quick_info.py"), and some don't (e.g. "jsonhelpers.py" or "eventhub.py").  It would be good to make this consistent while doing all the work to move files around now.  Personally, I like all lower case with underscores, e.g. "quick_info.py", which aligns with the PEP8 guidance. (p.s. "auto_indent_on_enter_between_curly_braces.py" is probably a little long.  Can we shorten this?)

Also, why did we make 'commands' and 'listeners' top level packages also, rather than residing within one top level package (e.g. 'libs')?

zhengbli:
I agree that all lower case with underscore file names would make more sense. 
I don't really have a preference regarding weather to make the 'commands' and 'listeners' top level packages or not, would that make a difference? Both anaconda and omnisharp are doing it too, so I didn't give it much thought.

billti:
It's not a big deal about the top-level thing, so we can leave it as is for now.  If everything was surfaced in the top-level package it would allow you to just have "from libs import *" in the typescript.py, but again, as-is currently is fine.

commands/__init__.py
```diff
@@ -0,0 +1,43 @@
+from .auto_indent_on_enter_between_curly_braces import TypescriptAutoIndentOnEnterBetweenCurlyBrackets
+from .error_info import TypescriptErrorInfo
+from .format import TypescriptFormatBrackets, TypescriptFormatDocument, TypescriptFormatLine, TypescriptFormatOnKey, \
```
billti: Several of these lines go way beyond the 79 char width recommended in PEP8.  You can use parens to group imports and not need the line continuation character.  For example, the above might look better as:

```python
from .format import (
    TypescriptFormatBrackets, 
    TypescriptFormatDocument, 
    TypescriptFormatLine, 
    TypescriptFormatOnKey,
    TypescriptFormatSelection, 
    TypescriptPasteAndFormat)
```

commands/auto_indent_on_enter_between_curly_braces.py
```diff
@@ -0,0 +1,25 @@
+from ..libs.viewhelpers import is_typescript, insert_text, set_caret_pos
+from .base_command import TypeScriptBaseTextCommand
+
+
+class TypescriptAutoIndentOnEnterBetweenCurlyBrackets(TypeScriptBaseTextCommand):
```
billti: This class name does not match the file name (*braces Vs *Brackets).

billti: Actually, can we just move this to format.py with the other format commands?

listeners/eventhub.py
```diff
@@ -0,0 +1,22 @@
...
+            cls.listener_dict[key] = [listener]
+
+    @classmethod
+    def run_listeners(cls, key, *args):
+        if key in cls.listener_dict.keys():
+            for handler in cls.listener_dict[key]:
+                handler(*args)
+
+    @classmethod
+    def run_listener_with_return(cls, key, *args):
```
billti: This doesn't seem right/safe to me.  What happens if multiple listeners sign-up for an event, and it is run via this method?  Only the first one will execute right?

ddotlic:
FYI: you broke a few things by renaming `jsonhelpers` to `json_helpers` but not updating all the places where this is used. Example: on lines 19 and 74 in commands/references.py you still call `jsonhelpers` but should call `json_helpers`, this breaks "Find All References"; I've fixed locally and verified that it works.

zhengbli:
@ddotlic Oops I just noticed as well. Will correct in a minute. Thanks for the heads up!

PR #202, opened 2015-05-20
Corrected naming for json_helpers module

zhengbli:
Fix the wrong renaming for json_helpers caused by refactoring.


billti:
:+1:

PR #205, opened 2015-05-22
Fix Sublime Text 2 issue after refactor

zhengbli:
The relative support in PR #197 didn't work with Sublime 2. Fix the importing issues and added a ``wait_count`` to prevent the infinite loop waiting for error info response during idle time.


TypeScript.py
```diff
@@ -1,19 +1,13 @@
...
-    from libs.view_helpers import *
-    from listeners import *
-    from commands import *
-else:
-    from .libs import *
-    from .libs.reference import *
-    from .libs.view_helpers import *
-    from .listeners import *
-    from .commands import *
+sys.path.append(os.path.dirname(os.path.abspath(__file__)))
```
billti: One of the goals of the refactor was to remove the need to modify the global module resolution path.  This is a bad practice, and we should avoid doing this, (and don't need to).  

We should only special case the top level package import (as you had in the previous refactor), and have everything else be intra-package.  (I use a similar pattern in a prototype at https://github.com/billti/BillsSTP/blob/master/init.py ).

zhengbli: I tried this approach but it didn't work. The separate treatment at the top level will indeed make top level import successful, but the cascading import later will still fail with "No module named typescript_module". So we either end up to do the relative importing in every file, or mess with the ``sys.path``.  Related links I found: 
http://stackoverflow.com/questions/9383014/cant-import-my-own-modules-in-python
http://stackoverflow.com/questions/24722212/python-cant-find-module-in-the-same-folder

TypeScript.py
```diff
@@ -1,19 +1,13 @@
...
+
+from typescript_module.libs import *
+from typescript_module.libs.reference import *
+from typescript_module.libs.view_helpers import *
+from typescript_module.listeners import *
+from typescript_module.commands import *
```
DanielRosenwasser: Not big on the name `typescript_module`; what about `typescript_sublime` or just `typescript`?

TypeScript.py
DanielRosenwasser: Not familiar with the above discussion, but if you do decide to continue using this approach, use [`os.path.join`](https://docs.python.org/2/library/os.path.html#os.path.join) instead of `append`.

typescript_module/listeners/idle.py
```diff
@@ -149,14 +151,21 @@ def on_idle(self):
         log.debug("call get_event from on_idle")
         ev = cli.service.get_event()
         if ev is not None:
-            self.dispatch_event(ev)
             self.error_refresh_requested = False
+            self.dispatch_event(ev)
             # reset the timer in case more events are on the queue
             self.set_on_idle_timer(50)
         elif self.error_refresh_requested:
+            # if the last request takes too long we'll abandon it
```
DanielRosenwasser: "drop" sounds a little less harsh than "abandon" :D

TypeScript.py
zhengbli: My intention was to avoid conflicts with the top level file name 'typescript.py'. But I guess we can name the file something like 'main.py' and name the module 'typescript'

typescript_module/listeners/idle.py
```diff
@@ -149,14 +151,21 @@ def on_idle(self):
...
+            # to prevent infinite loop
```
DanielRosenwasser: to prevent an infinite loop

TypeScript.py
zhengbli: ``os.path.join`` seems to be doing different things than ``sys.path.append``? The latter is to add the path to the known list where python searches for modules.

billti: Back to the original topic... those articles don't disagree with my comment, and my pattern is working fine for me in ST2 & ST3.  Let's discuss in person on the issue and how it should work.

zhengbli: Given it another try, now it indeed worked. I think the reason is that last time I tried I didn't put everything under the same root folder. I'll test a bit more and update soon.

typescript/libs/global_vars.py
```diff
@@ -2,13 +2,14 @@
 import re
 
 import sublime
+from os.path import dirname
 from .logger import *
```
billti: There seems to be a circular reference here.  global_vars imports from logger, yet logger imports from global_vars, and tries to use values defined in it.

typescript/libs/global_vars.py
```diff
@@ -2,13 +2,14 @@
...
 
 
 # get the directory path to this file; 
-LIBS_DIR = os.path.dirname(os.path.abspath(__file__))
-PLUGIN_DIR = os.path.dirname(LIBS_DIR)
-PACKAGES_DIR = os.path.dirname(PLUGIN_DIR)
+LIBS_DIR = dirname(dirname(os.path.abspath(__file__)))
```
billti: This is confusing, as you have a 'libs' dir (i.e. that this file is in), yet this folder points to '/Users/bill/src/TypeScript-Sublime-Plugin/typescript' on my machine (i.e. the package dir).  Is this just named confusingly?

typescript/libs/global_vars.py
```diff
@@ -2,13 +2,14 @@
...
+PLUGIN_DIR = dirname(LIBS_DIR)
+PACKAGES_DIR = dirname(PLUGIN_DIR)
```
billti: This doesn't handle if my plugin is installed via a symlink. For example, on my machine this points to '/Users/bill/src' - which is clearly not what is intended.  Why not just use `sublime.packages_path()`, which on my machine is defined as `/Users/bill/Library/Application Support/Sublime Text 2/Packages`

typescript/libs/global_vars.py
zhengbli: I wonder why it didn't give me an error... But you are right, better change it to ``import logging`` to avoid future problems.

typescript/libs/global_vars.py
zhengbli: Yes it is a naming confusion. I'm thinking about changing to ``PACKAGE_DIR``.

billti:
Regardless of the above comments, the functionality is looking good for me on ST2 and ST3 on OS X.  I'd recommend merging it in so we have a base to keep improving on.  Thanks! :+1:

typescript/libs/global_vars.py
zhengbli: The reason is ``sublime.packages_path()`` returns nothing if run at importing time. And the vars are used in some places during importing time. Alternative solution is to initialize them during ``plugin_loaded`` function, which is a problem for sublime 2 because it is not called in sublime 2. Therefore I figure it would be easier if I can figure the path out during importing time. I'll explore a bit to see how python deal with symlink on linux and OS X, it looks it behaves differently from windows.

billti: Correct, the Sublime APIs are available immediately on ST2, but not until plugin_loaded on ST3.  I used a pattern at https://github.com/billti/BillsSTP/blob/master/init.py#L14 which worked pretty well, but doing work in a init function that is called on load in ST2, and in plugin_loaded on ST3.

Generally we want to avoid work in module loading anyway where possible as a best practice.

On a meta-note, if this path is not reliable, what are we using it for?  Do we need it at all?

typescript/libs/global_vars.py
billti: That doesn't solve anything.  You have code that runs at module load time in both modules.  It doesn't matter how/what you load from the module, when you load it, all the global code runs.

zhengbli: I mean ``logging`` as the global python module, I wasn't referring to the ``logger`` file in this plugin. That was the only thing used in the ``global_vars.py`` anyways.

typescript/libs/global_vars.py
zhengbli: We use it mainly to load resources, like the temporary files for reloading and html template files for tooltips. I think the pattern you mentioned is ideal for retrieving the sublime package path, but we do need to get the plugin's path also, because we load binary sources from within the plugin folder and the path has to have a format like ``Packages/[PLUGIN_NAME]/icons``.

Issue #194, opened 2015-05-12
semicolon stops working


jamiecounsell:
When using the plugin, the semicolon stops working with no indication of why and no pattern. Restarting Sublime Text (3083) did not work for me, but toggling `typescript_auto_format` did.

zhengbli:
Do you install from package control or cloned the master branch from the repo?

aMarCruz:
Same issue here.
Win7 32bits, Typescript 1.5b, Sublime Text 3 build 3083. ts plugin 1.4 installed from package control.

zhengbli:
@jamiecounsell @aMarCruz can you paste the log output in your sublime console? Or better, can you do the following:
1. Go to the typescript plugin folder, edit the ``TypeScript.py`` file to make ``logFileLevel = logging.DEBUG``
2. Open sublime to reproduce the problem, then close it.
3. Open the ``TS.log`` file in the plugin folder (using other editors), paste the content here.
It could be due to that the node server hasn't started.

aicioara:
@jamiecounsell how do you toggle `typescript_auto_format` ? 
There is also a related issue, which might be more relevant. Having multiple cursors, hitting semicolon, only inserts semicolon on the first cursor. I believe it has to do because semicolon auto formats the enclosing line(s) and it stops propagating the event to the rest of the cursors. However, I don't want auto format of the text anyway, I can type properly, how do I turn it off?

jamiecounsell:
@zhengbli I can't replicate the issue right now. If it happens again I will post the log.
@aicioara glad you figured it out. That's how I did it too. Yes, that would be nice. If it happens again, follow @zhengbli and get the log. Better yet, maybe turn on the log all the time anyway! Hopefully we can get some more information to resolve this issue.

zhengbli:
It seems that the way we trigger ``typescript_format_on_key`` command is breaking a range of default editing behaviors in sublime editor, like multi-cursor etc, because we interfere the text inserting process within the command. I'm considering move all the triggering to the ``on_modified`` listener, which in the worst case will guarantee the characters be shown.

mdausmann:
+1 this is infuriating! also thanks @aicioara the setting change works.  BTW, are semicolons considered 'bad' syntax for typescript, or this is just a bug?

zhengbli:
[...]

PR #209, opened 2015-05-28
Move the trigger of ``format_on_key`` command to ``on_modified``

zhengbli:
Our current trigger for the ``format_on_key`` command is to intercept the keystroke before it appears in the buffer, which has caused many problems:
1. The typed characters wouldn't show up if the command failed;
2. For multiple cursors, only the first cursor would insert test because we choose to do the insertion by ourselves;
3. Lag before the characters show up if the formatting takes too long.

Therefore it might be better to move the trigger for ``format_on_key`` to the ``on_modified`` event listener, because the listener is called **after** the characters are inserted. 

Fix issue: #194

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -135,57 +135,6 @@
             { "key": "selector", "operator": "equal", "operand": "source.ts" },
             { "key": "is_popup_visible"}
         ]
-    },    // In case when auto match is enabled, only format if not within {}
-    {
-        "keys": [ "}" ],
-        "command": "typescript_format_on_key",
-        "args": { "key": "}" },
-        "context": [
-            { "key": "selector", "operator": "equal", "operand": "source.ts" },
-            { "key": "setting.typescript_auto_format", "operator": "equal", "operand": true },
-            { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
-            { "key": "num_selections", "operator": "equal", "operand": 1},
-            { "key": "selection_empty", "operator": "equal", "operand": true },
-
-            { "key": "preceding_text", "operator": "not_regex_contains", "operand": "\\{$" },
-            { "key": "following_text", "operator": "not_regex_contains", "operand": "^\\}" }
-        ]
-    },
-    // In case when auto match is disabled, format the block regardless
-    {
-        "keys": [ "}" ],
-        "command": "typescript_format_on_key",
-        "args": { "key": "}" },
-        "context": [
-            { "key": "num_selections", "operator": "equal", "operand": 1},
-            { "key": "setting.typescript_auto_format", "operator": "equal", "operand": true },
-            { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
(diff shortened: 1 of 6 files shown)


typescript/listeners/format.py
```diff
@@ -11,5 +13,35 @@ def on_post_text_command_with_info(self, view, command_name, args, info):
...
+            last_command, args, repeat_times = view.command_history(0)
+            log.debug("last_command:{0}, args:{1}".format(last_command, args))
+            if last_command == "insert":
+                pos = info.prev_sel[0].begin()
+                if ";" in args["characters"]:
+                    view.run_command("typescript_format_on_key", {"key": ";"})
+                if "}" in args["characters"]:
+                    if cli.auto_match_enabled:
+                        prev_char = view.substr(pos - 1)
+                        post_char = view.substr(pos + 1)
```
billti: Is this (and the command above) save it you're at the start or end of the file? (i.e. you don't get an index out of range exception or similar?

Default.sublime-keymap
```diff
@@ -135,57 +135,6 @@
...
-            { "key": "selector", "operator": "equal", "operand": "source.ts" }
-        ]
-    },
-    {
-        "keys": [ "enter" ],
-        "command": "typescript_format_on_key",
-        "args": { "key": "\n" },
-        "context": [
-            { "key": "setting.typescript_auto_indent", "operator": "equal", "operand": true },
-            { "key": "auto_complete_visible", "operator": "equal", "operand": false },
```
billti: I see a bunch of settings and checks here I don't see repeated in the imperative logic below. Was there a need for them?  Does it still behave correctly without them?

typescript/listeners/format.py
zhengbli: if you are at point before the beginning of the buffer or after the ending of the buffer, ``view.substr`` won't throw exception, it will simply return `\x00`. Which is fine because they wouldn't be equal to `{` or `}` to interfere the conditions.

Default.sublime-keymap
zhengbli: Most of the conditions are incorporated: 
- ``selection_empty == True`` -> ``info.prev_sel[0].empty()``
- ``setting.typescript_auto_indent`` -> ``cli.ts_auto_indent_enabled``
- ``num_selections == 1`` -> ``len(info.prev_sel) == 1``
- ``selector contains 'meta.scope.between-tag-pair' `` -> ``view.score_selector(pos, "meta.scope.between-tag-pair") > 0``
etc.

The only change I made was regarding the ``auto_complete_visible``, I am not sure we need that condition. To be more specific, I think we still need to do the formatting if the user types an `;` when the auto completion list is up.

billti:
:+1:

zhengbli:
@billti I found out that the plugin path was still incorrect for Sublime Text 2 on mac if the plugin is installed as a file link; and it has impact when I tried 'find all reference' that it cannot find the syntax definition file (because it was searching in the wrong place). So I added a fix in this PR in commit 85f511f.

PR #210, opened 2015-05-28
Add build module

zhengbli:
Adding support for building loose ts files & projects with ``tsconfig.json``.
~~Note: on mac it is required to install the [``Fix Mac Path``](https://github.com/int3h/SublimeFixMacPath) for the build command to work. Otherwise the build command will not be found.~~

Behavior:
The build command is invoked by ``ctrl + b``, or by menu item ``Tools -> Build``
- For loose ts files:
  - First an input panel will pop up for the user to input [compiling options](https://github.com/Microsoft/TypeScript/wiki/Compiler-Options)
  - Then the build result will show up in an output panel.
- For project files with ``tsconfig.json`` file, the command will load the ``tsconfig`` file and compile accordingly. 

Demo:
test


typescript/commands/build.py
```diff
@@ -0,0 +1,40 @@
...
+import sublime
+import os
+from ..libs.global_vars import IS_ST2
+
+
+class TypescriptBuildCommand(sublime_plugin.WindowCommand):
+    def run(self):
+        file_name = self.window.active_view().file_name()
+        directory = os.path.dirname(file_name)
+        if "tsconfig.json" in os.listdir(directory):
```
billti: Does this detect a tsconfig.json in any parent dir?  That's what the server does, and uses that for the language service settings.  If the server is already looking for and loading this file, maybe better to just ask the server if it knows about this file and where, than repeat logic to locate it.

billti:
Great!  I'd rather not have the dependency on another package (`Fix Mac Path`).  Could we attempt to locate TSC/Node based on the locations it's commonly installed by the installer or brew?  Or at least (or also) allow the user to add a setting for the location to use at build time?

typescript/commands/build.py
zhengbli: There is no existing command on the server side to provide project info. However, I do think we need to create one for not only this and future features like project-wide error listing etc. I'll send a PR to the typescript repo soon

typescript/commands/build.py
```diff
@@ -0,0 +1,40 @@
...
+            self.window.run_command("exec", {
+                "cmd": ["tsc"],
```
billti: I thought we were going to use the TypeScript compiler that effectively ships with the Sublime plugin (i.e. is included in tsserver.js).  Otherwise if you have a different version of the TypeScript Node package installed than the plug-in version, you might get different errors/features from your build than you do from the language service while editing - which would be weird.

As we're already locating Node and tsserver.js to spin up the plugin service, it shouldn't be too hard to use the same located files to run the compilation for the build.  What do you think?

zhengbli: Hmmm, that's very good point. I need to understand more about how this could be done, but I agree it is indeed important to keep the versions consistent. I'll try that direction.

zhengbli:
Now the command works as long as the user has node installed (no global tsc requirement any more), and it retrieves tsconfig file from the server instead of search on its own. Tested on mac and windows.

typescript/commands/build.py
```diff
@@ -0,0 +1,40 @@
+import sublime_plugin
+
+from ..libs.global_vars import *
+from ..libs import cli
+
+
+class TypescriptBuildCommand(sublime_plugin.WindowCommand):
+    def run(self):
+        if get_node_path() is None:
+            print("Cannot found node. Build cancelled.")
```
billti: * found -> find.  Should probably be "Cannot find NodeJS".

typescript/commands/build.py
```diff
@@ -0,0 +1,40 @@
...
+                    "Build parameters: ",
+                    "",  # initial text
+                    self.compile_inferred_project,
+                    None,  # on change
+                    None   # on cancel
+                )
+
+    def compile_inferred_project(self, params=""):
+        file_name = self.window.active_view().file_name()
+        cmd = [get_node_path(), TSC_PATH, file_name]
```
billti: Can't you just pass this in from line 13 above via a lambda?  Seems awkward to regrab it at this point (e.g. is it possible the focus changed since they triggered the build command)?

typescript/commands/build.py
```diff
@@ -0,0 +1,40 @@
...
+        print(cmd)
+        if params != "":
+            cmd.extend(params.split(' '))
+        self.window.run_command("exec", {
+            "cmd": cmd,
+            "file_regex": "^(.+?)\\((\\d+),(\\d+)\\): (.+)$"
```
billti: What is the purpose of this file_regex param?  You are already passing the file to compile as part of the 'cmd' arg.  If this is needed, please comment why, as the usage isn't clear.

zhengbli: The ``file_regex`` is for capturing the build result output that we are interested, and display them in the build output panel.

billti:
:+1:

Issue #201, opened 2015-05-20
Auto-completion of statement by pressing Tab is not aware of surrounding code


v95:
If I am typing `console.l` and hit the `tab` button (once the auto-complete dropdown appears), instead of simply adding `og();`, the line becomes the following:

`console.console.log();`

I am using Sublime Text 2 on Win10. Please let me know if more information is required.

zhengbli:
I don't seem to repro problem on Sublime Text 2 with v1.4 of the plugin (from package control):
![capture](https://cloud.githubusercontent.com/assets/1171301/7781197/099993c0-009f-11e5-957f-3f0f3c7451c8.gif)

Notice a recent refactor change (#197) broke the support for sublime text 2 (sorry about that). If you want to install from the git repo, please wait after the fix (#205) to merge.

zhengbli:
The same with #211 
The problem is sometimes snippet is inserted instead of the completion information. Working on a fix.

v95:
Thanks @zhengbli!

PR #214, opened 2015-06-05
Fix confusing snippet triggers

### LINKED ISSUE #211: Autocomplete fills in complete text even if a part of it is already written.
The autocomplete feature present in the TypeScript plugin automatically puts in the complete text that the user actually wants to write, even if a part of the text is already written, a small example is as follows

### Steps to Reproduce

```ts
export function addPin(userid: string, boardid: string, imageUri: string, link: string, caption: string, callback: (user: User) => void) {
    db.collection('images', function(error, images_collection) {
        if(error) { console.error(error); return; }
        images_collection.insert({
            user: userid,
            caption: caption,
            imageUri: imageUri,
            link: link,
            board: boardid,
            comments: []
        }, function(error, image) {
            console.log(image);
            db.collection('users', function(error, users) {
                if(error) { console.error(error); return; }
                users.update(
                    {_id: userid, "boards.title": boardid}, 
                    {"$push": {"boards.$.images": image[0]._id}},
                    function(error, user) {
                        callback(user);
                    }
                );
            })
        })
    })
}

```
1. Take the following code example
2. Try to add a `console.log()`
3. As soon as the user types the word `console.l` the <log()> appears in the dropdown autocomplete options. Hitting enter results in `console.console.log()` rather than `console.log()`
4. Simply using `log()` and hitting enter on the autocomplete options results in `console.log()`

### Expected Results

The part of the string that is already written shouldn't be repeated again in the auto completion option just like how it happens in autocomplete options for other languages like JavaScript where the above similar steps just result in `console.log()` than a `console.console.log()`

### Actual Results

`console.console.log()` instead of `console.log()`

zhengbli:
Some of the snippet triggers have the same name with common TS method/property names; it can lead to confusing behavior when they show up together in the completion list. For example, when typing ``console.`` both the snippet ``log`` and the member ``log`` would show up, but they will insert different content once committed. 

Fix #211 and #201

### DIFF
--- a/snippets/get-()-{}.sublime-snippet
+++ b/snippets/get-()-{}.sublime-snippet
@@ -4,7 +4,7 @@ public get ${1:value}() : ${2:string} {
 	${3:return }
 }
 ]]></content>
-	<tabTrigger>get</tabTrigger>
+	<tabTrigger>getter</tabTrigger>
 	<scope>source.ts</scope>
     <description>get-property …</description>
 </snippet>
--- a/snippets/log.sublime-snippet
+++ b/snippets/log.sublime-snippet
@@ -1,7 +1,7 @@
 <snippet>
     <content><![CDATA[console.log($1);
 $0]]></content>
-    <tabTrigger>log</tabTrigger>
+    <tabTrigger>clog</tabTrigger>
     <scope>source.ts</scope>
     <description>console log</description>
 </snippet>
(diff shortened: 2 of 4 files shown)


zhengbli:
@DanielRosenwasser could you take a look at the new triggers for the snippets? It should be a trivial fix. Thanks!

billti:
:+1:

Issue #208, opened 2015-05-27
Generic expressions in class body are highlighted incorrectly


sccolbert:
The attached image demonstrates the problem. When an expression in the class body has generic type parameters, it is highlighted incorrectly. The top two lines have the correct highlighting.

![badhighlight](https://cloud.githubusercontent.com/assets/137289/7843550/48bcb3d4-0478-11e5-9af1-26c2f65288b9.PNG)

sccolbert:
cheers!

Issue #191, opened 2015-05-12
Syntax Highlighting Bug


musicist288:
There seems to be an issue with the syntax highlighting, at least inside the constructor. Looks like it's intreperting the `/*` sequence inside the string literal `"**/*.json"` as the start of a comment block. 

![syntax-highlighting](https://cloud.githubusercontent.com/assets/368976/7588809/5caf4d26-f88d-11e4-9d96-e9c475a99afa.PNG)

zhengbli:
I don't seem to reproduce the problem (latest bit in master):
![capture](https://cloud.githubusercontent.com/assets/1171301/8016748/8242e256-0b9c-11e5-8bbd-91b4f0a7227e.PNG)

Can you give it another try? If it still occurs, can you provide a bit more about your code, like what is the outer scope and other things may interfere with the highlighting?

musicist288:
It's still happening for me. Though I figured out a very small change that causes it. 

It looks like it's broken when the opening curly brace is on a new line instead of next to the function name and arguments.

Broken:
![broken](https://cloud.githubusercontent.com/assets/368976/8035381/12f55eb4-0dbe-11e5-974b-5ecbbbc38c7b.PNG)

Working:
![working](https://cloud.githubusercontent.com/assets/368976/8035383/190c41be-0dbe-11e5-92f0-616c48f72d14.png)

zhengbli:
Can you give #215 a try? It should be fixed there.

musicist288:
Works for me. Thanks!

zhengbli:
Great!

PR #215, opened 2015-06-06
Fix highlighting issues for constructors containing type parameter

zhengbli:
Fix issue #208 and #191


TypeScript.YAML-tmLanguage
```diff
@@ -309,9 +318,10 @@ repository:
 
   type-parameters:
     name: meta.type.parameters.ts
-    begin: '<'
+    begin: '(\w+)?(<)'
```
hoanhtien: Should <code>([a-zA-Z_$][\w$]*)?(<)</code> be more accurate?

zhengbli: Fixed

hoanhtien:
+1

Issue #216, opened 2015-06-06
Performance issues on Angular code base


mprobst:
I'm using Sublime Text 3, build 3083, on the Angular code base, with the TypeScript plugin installed via git at revision 2c19761, tried with node v0.10.38 and v0.12.4 with similar results.

At its worst, moving the cursor stalls the UI for multiple seconds and most TypeScript commands fail (go to symbol, rename, etc). I see "queue timeout" printed in the console. Restarting Sublime helps, but moving the cursor onto a symbol still stalls the UI for a noticeable amount of time; keeping the down arrow pressed and scrolling through the file shows it's very slow and janky.

Steps to reproduce:
* install Sublime Text 3 build 3083
* install TypeScript plugin at 2c19761
* Check out https://github.com/angular/angular (about 60k SLOC)
* open (e.g.) element_injector.ts, navigate around in the file

Apart from the general speed, all TypeScript operations appear to run synchronous on the UI thread, which might cause these issues.

zhengbli:
Just to clarify, when you say "moving the cursor onto a symbol", do you mean just click on the symbol, or do you mean using the "navigate to symbol" command? 
Some commands are kept synchronous due to comparability consideration with sublime 2, which may indeed affect the performance.  However, the expensive ones like auto_completion and get_err are done asynchronous in Sublime 3 already.

mprobst:
Just clicking on it or navigating by keyboard.

Zhengbo Li <notifications@github.com> schrieb am Fr., 5. Juni 2015 19:17:

zhengbli:
Does the following recording seem to match your experience? I tried the procedure on my laptop:
![capture](https://cloud.githubusercontent.com/assets/1171301/8017889/b6ac6e4c-0bb8-11e5-8c92-93fce6fe043f.gif)
I did notice some lag, but no as bad. I wonder there might be something else going wrong. Could you help me with a log file? To get the log file:
1. go to the plugin folder and close sublime
2. open 'typescript/libs/global_vars', change ``LOG_FILE_LEVEL`` to ``logging.DEBUG``
3. open sublime and reproduce the problem.
4. find the 'TS.log' file in the 'typescript' folder, and post the content.

Thanks!

mprobst:
[...]

PR #217, opened 2015-06-06
Change quick info request to async for ST3

zhengbli:
Fix issue #216

### DIFF
--- a/typescript/libs/node_client.py
+++ b/typescript/libs/node_client.py
@@ -25,7 +25,7 @@ class CommClient:
 
     def sendCmdSync(self, cmd): pass
 
-    def sendCmdAsync(self, cmd): pass
+    def sendCmdAsync(self, cmd, cb): pass
 
 
 class NodeCommClient(CommClient):
@@ -111,7 +111,7 @@ class NodeCommClient(CommClient):
         }
         return timeoutMsg
 
-    def sendCmd(self, cmd, seq, cb):
+    def sendCmd(self, cmd, cb, seq):
         """
         send single-line command string; no sequence number; wait for response
         this assumes stdin/stdout; for TCP, need to add correlation with sequence numbers
@@ -133,7 +133,7 @@ class NodeCommClient(CommClient):
             if (cb):
                 cb(self.makeTimeoutMsg(cmd, seq))
 
-    def sendCmdAsync(self, cmd, seq, cb):
+    def sendCmdAsync(self, cmd, cb, seq):
         """
         Sends the command and registers a callback
         """
--- a/typescript/libs/node_client.py
+++ b/typescript/libs/node_client.py
@@ -21,7 +21,7 @@ class CommClient:
 
     def postCmd(self, cmd): pass
 
-    def sendCmd(self, cb, cmd): pass
+    def sendCmd(self, cmd, cb): pass
 
     def sendCmdSync(self, cmd): pass
 
@@ -111,7 +111,7 @@ class NodeCommClient(CommClient):
         }
         return timeoutMsg
 
-    def sendCmd(self, cb, cmd, seq):
+    def sendCmd(self, cmd, seq, cb):
         """
         send single-line command string; no sequence number; wait for response
         this assumes stdin/stdout; for TCP, need to add correlation with sequence numbers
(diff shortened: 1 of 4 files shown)


typescript/libs/service_proxy.py
```diff
@@ -155,11 +156,18 @@ def quick_info(self, path, location=Location(1, 1), onCompleted=None):
...
-            req_dict["seq"]
-        )
+        if not IS_ST2:
+            self.__comm.sendCmdAsync(
+                json_str,
+                req_dict["seq"],
+                lambda json_dict: None if onCompleted is None else onCompleted(json_dict)
+            )
+        else:
+            self.__comm.sendCmd(
```
DanielRosenwasser: Is there any reason why `sendCmdAsync` and `sendCmd` don't take arguments in the same order?

typescript/libs/service_proxy.py
```diff
@@ -155,11 +156,18 @@ def quick_info(self, path, location=Location(1, 1), onCompleted=None):
...
-        self.__comm.sendCmd(
-            lambda json_dict: None if onCompleted is None else onCompleted(json_dict),
-            json_str,
-            req_dict["seq"]
-        )
+        if not IS_ST2:
+            self.__comm.sendCmdAsync(
+                json_str,
+                req_dict["seq"],
+                lambda json_dict: None if onCompleted is None else onCompleted(json_dict)
```
DanielRosenwasser: Factor the callback out so that you're not duplicating code.

DanielRosenwasser: Also consider making this `onCompleted and onCompleted(json_dict)`.

Though I'm not sure how idiomatic that is in Python.

typescript/libs/service_proxy.py
zhengbli: It just happened to be that way, unifying it.

typescript/libs/service_proxy.py
```diff
@@ -155,11 +156,19 @@ def quick_info(self, path, location=Location(1, 1), onCompleted=None):
         args = {"file": path, "line": location.line, "offset": location.offset}
         req_dict = self.create_req_dict("quickinfo", args)
         json_str = json_helpers.encode(req_dict)
-        self.__comm.sendCmd(
-            lambda json_dict: None if onCompleted is None else onCompleted(json_dict),
-            json_str,
-            req_dict["seq"]
-        )
+        callback = lambda json_dict: None if onCompleted is None else onCompleted(json_dict)
```
DanielRosenwasser: What about just passing in `onCompleted or lambda: None` here as well as in all other locations?

typescript/libs/service_proxy.py
```diff
@@ -45,9 +46,9 @@ def completions(self, path, location=Location(1, 1), prefix="", on_completed=Non
         req_dict = self.create_req_dict("completions", args)
         json_str = json_helpers.encode(req_dict)
         self.__comm.sendCmd(
-            lambda json_dict: None if on_completed is None else on_completed(json_dict),
             json_str,
-            req_dict["seq"]
+            req_dict["seq"],
+            lambda json_dict: None if on_completed is None else on_completed(json_dict)
```
DanielRosenwasser: Naming here isn't consistent.

typescript/libs/service_proxy.py
zhengbli: I didn't quite understand last time you mentioned it. Now I got it :). That would be more concise.

DanielRosenwasser: That's okay, the last thing I suggested was a lot stupider :smile:

DanielRosenwasser:
:+1:

typescript/libs/node_client.py
```diff
@@ -111,7 +111,7 @@ def makeTimeoutMsg(self, cmd, seq):
         }
         return timeoutMsg
 
-    def sendCmd(self, cb, cmd, seq):
+    def sendCmd(self, cmd, seq, cb):
```
billti: You have derived a class from CommClient and overridden this sendCmd method, you have have changed the meaning of the first 3 params (which are "self, cmd, cb" in the base class).  You should keep the matching parameters the same in the derived class (as they were in the original code).

zhengbli: Changed that to match the same order. The original code seemed wrong too, as `seq` appears in the middle. I think it won't affect the functionality as the number of parameters is different, but I agree for consistence they should be kept the same order.

PR #222, opened 2015-06-08
Fix the issue that tooltip doesn't have proper style in Windows 64 version ST 3

### LINKED ISSUE #219: No Style for tooltip (master) & Font size in the tooltip (v0.1.4)
Hi,

just realized that with the latest official released version (0.1.4) the tooltips (which I've had for quite some time because I used the HEAD version of the master from github repo) are finally colorized (this did not work - in the sense that popup was black text on white - before and still doesn't if I just clone the repo, weird).

However, the font size is a bit small, let me try to demonstrate:

fontsize

Is there a way to configure the font size?

zhengbli:
The reason the styles cannot be applied seems to be the CR line endings in the html string. Proper style is working after removing the line endings.

Fix #219

### DIFF
--- a/typescript/libs/popup_manager.py
+++ b/typescript/libs/popup_manager.py
@@ -1,3 +1,5 @@
+import re
+
 from string import Template
 
 from .logger import log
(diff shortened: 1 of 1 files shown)


billti:
Bizarre cause, but thanks for fixing! :+1:

trubit:
Does not seems to fix this problem. Win64, ST3 3114, plugin v. 0.1.15
I still see unstyled popup.

zhengbli:
This is because of the change in #511 so that we don't hardcode the style and leave it for the themes to define the styling. Though we can offer a default styling somehow.

PR #224, opened 2015-06-10
Remove insert_key parameter

steveluc:
Removed obsolete insert_key parameter from format_on_key.  Re-enabled format line and format brackets, which were broken due to rename of insert key parameter.


zhengbli:
:+1:

orta:
Hi! Blast from the past, I'm struggling to reproduce this (even though the code still has the `insert_key` field. I'm going to close this, if someone can show me how to repro it - I'll re-open and get the now rebased PR back.

PR #236, opened 2015-06-15
Display last build parameters in build dialog.

hoanhtien:


typescript/commands/build.py
```diff
@@ -23,7 +25,7 @@ def run(self):
             else:
                 sublime.active_window().show_input_panel(
                     "Build parameters: ",
-                    "",  # initial text
+                    self.build_parameters,
```
zhengbli: isn't ``build_parameters`` a class property instead of an instance property? I thought it can only be referenced by ``TypescriptBuildCommand.build_parameters``

typescript/commands/build.py
```diff
@@ -38,3 +40,4 @@ def compile_inferred_project(self, file_name, params=""):
             "cmd": cmd,
             "file_regex": "^(.+?)\\((\\d+),(\\d+)\\): (.+)$"
         })
+        self.build_parameters = params
```
zhengbli: What happens if you don't give any parameter, will the ``params`` be ``None`` or empty string?

typescript/commands/build.py
zhengbli: Also it might be better to keep the comment of the argument

hoanhtien: Yes you are right. Fixing...

typescript/commands/build.py
hoanhtien: I think empty string as <code>params=""</code> means the default value of <code>params</code> is empty string.

zhengbli: On second look, this might not be working, as you didn't pass the saved ``build_parameters`` to the ``compile_inferred_project`` callback. The ``params`` in the lambda expression means differently from the ``params`` elsewhere because they have different scope. Does this work on your machine?

zhengbli: I mean if it is OK to pass ``None`` as the second argument to ``show_input_panel`` function. Also, does this change work on both sublime 2 and 3? We need to test for both

zhengbli:
:+1:

Issue #226, opened 2015-06-12
Comment not colored correctly


dbaeumer:
Some examples:
```
var x = new /*asdasd*/ String();

interface /*asdasd*/ A {
   a?:string;
}
```
From looking at the grammer I think there are more places where comments aren't colored correctly

hoanhtien:
The colorization issue in `var x = new /*asdasd*/ String();` is fixed in https://github.com/Microsoft/TypeScript-Sublime-Plugin/pull/238.

hoanhtien:
Updated #238 to fix `interface /*asdasd*/ A` as well.

PR #238, opened 2015-06-17
Fix common colorization issue in class/interface declaration.

hoanhtien:
Before:
image

After:
image


TypeScript.YAML-tmLanguage
```diff
@@ -76,10 +76,21 @@ repository:
...
+    - include: '#object-heritage-parent'
+    - include: '#comment'
+  
+  object-heritage-parent:
+    name: meta.object.heritage.parent.ts
+    match: '(?:\s*([a-zA-Z_$][\w$]*))'
     captures:
-       '1': { name: keyword.other.ts }
-       '2': { name: storage.type.ts } 
+      '1': { name: storage.type.ts }
```
zhengbli: Does this cover the following case (from [ts spec](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md)):
```
interface KeyValuePair<K, V> extends Array<K | V> { 0: K; 1: V; }
```

hoanhtien: There is a small problem with type argument i.e. <code>\<K | V\></code> in the example above. This is fixed in the commit below.

hoanhtien:
The new commit fixes all scenarios in https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/226.

zhengbli:
:+1:

Issue #228, opened 2015-06-12
String expressions in field initializer not colored correctly if spawn multiple lines


dbaeumer:
```
class MyClass {
    template = '' +
               '<a> {} </a>';
}
```

zhengbli:
This should be fixed by #239 already.

PR #239, opened 2015-06-17
Extend scope of meta.field.declaration.ts to next lines by removing \n from the end mark

hoanhtien:
This should fix issue: [228](https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/228)

Before:
image

After:
image


zhengbli:
Will this affect method overloads that don't have semicolons?

e.g.
```typescript
class MyClass {
    foo(pa: number): () => number
    foo(pa: number)
    foo(pa: any) {
    }
}
```

hoanhtien:
It won't because I only make changes in meta.field.declaration.ts - should not affect meta.method.declaration.ts.

PR #242, opened 2015-06-18
Add .tsx to supported file extensions

RyanCavanaugh:


RyanCavanaugh:
:heart: @msftclas

TypeScript.tmLanguage
```diff
@@ -5,6 +5,7 @@
 	<key>fileTypes</key>
 	<array>
 		<string>ts</string>
+		<string>tsx</string>
```
zhengbli: This .tmLanguage file is generated, so changes will be overwritten next time it is generated. The source is in ``TypeScript.YAML-tmLanguage``. You need to change the ``fileTypes`` there, and use the [AAA package](https://packagecontrol.io/packages/AAAPackageDev) to generate the tmLanguage file.

Issue #187, opened 2015-05-08
Settings change event triggering unnecesary calls to reconfig_file


jgoz:
My setup:

- Sublime 3085 (Mac and Windows)
- [Vintageous](https://github.com/guillermooo/Vintageous)
- TypeScript-Sublime-Plugin (master)

I can't determine what is causing this behaviour, but it may be a bug in the Sublime settings API.

Basically, on every cursor movement in *normal* mode in Vintageous, the `tab_size_changed` event gets triggered, which then calls `reconfig_file` and issues a `configure` call to the TS service. This makes the editing experience *extremely* sluggish. Note that this does not happen in insert mode.

I did some tracing and came up with this:

```
  File "C:\Program Files\Sublime Text 3\sublime_plugin.py", line 533, in run_
    return self.run(**args)
  File "xactions in C:\Users\johng\AppData\Roaming\Sublime Text 3\Installed Packages\Vintageous.sublime-package", line 791, in run
  File "state in C:\Users\johng\AppData\Roaming\Sublime Text 3\Installed Packages\Vintageous.sublime-package", line 345, in sequence
  File "vi.settings in C:\Users\johng\AppData\Roaming\Sublime Text 3\Installed Packages\Vintageous.sublime-package", line 244, in __setitem__
  File "C:\Program Files\Sublime Text 3\sublime.py", line 982, in set
    sublime_api.settings_set(self.settings_id, key, value)
  File "C:\Users\johng\AppData\Roaming\Sublime Text 3\Packages\TypeScript\TypeScript.py", line 458, in tab_size_changed
    reconfig_file(view)
  File "C:\Users\johng\AppData\Roaming\Sublime Text 3\Packages\TypeScript\TypeScript.py", line 450, in reconfig_file
    cli.service.configure(hostInfo, view.file_name(), formatOptions)
  File "C:\Users\johng\AppData\Roaming\Sublime Text 3\Packages\TypeScript\libs\serviceproxy.py", line 28, in configure
    print(''.join(traceback.format_stack()))

DEBUG:TS:Posting command: {"seq": 14, "type": "request", "arguments": {"hostInfo": "Sublime Text version 3085", "file": "C:\\Code\\web-mapping\\MapViewModel.ts", "formatOptions": {"indentSize": 4, "tabSize": 4, "convertTabsToSpaces": true}}, "command": "configure"}
```

That command gets posted multiple times for every cursor movement in normal mode.

[...]

PR #243, opened 2015-06-19
Project Error List

zhengbli:
Add the project error list feature. The error list is a panel that displays all the errors in the current project. The list is updated in real time. The current project can be either a configured project with a ``tsconfig.json`` file, or a inferred project for a loose file. 

Because node is single threaded and analyzing errors in the project level constantly is expensive, we created a dedicated worker server just for the task. To avoid taking up too much memory, the serve only starts if the error list panel is active, and it will be shutdown once the panel is closed. 

Demo:
errorlist

Note in the task manager it can be seen how the worker node server is created and closed.

Note2: this feature is only available for ST3 due to the multi-threading issues in ST 2.


johnfn:
This looks really awesome! However, when I started testing it out, it would cause Sublime 3 to periodically freeze about every 20 seconds on a relatively small project (couple hundred lines of TS). I'm running OSX Yosemite, and my machine is a few years old but it's still pretty fast. I do think it's coming from this branch because I switched back to master and the behavior disappeared. 

The weirdest thing is that the errorlist does not even need to be open for this to happen. I caused it simply by checking out the branch and typing in some TypeScript.

zhengbli:
I did change the logic for the error requests and quick infos. However it seemed to run fine on my old Mac mini (Yosemite) when I tested it. What do you see in the console output? Could you change the log level to DEBUG and paste the output here? Thanks for testing it btw

johnfn:
Sure! 

https://gist.github.com/johnfn/0d8d1bf7e4de78d0e4bc
Here's another one: https://gist.github.com/johnfn/0d8d1bf7e4de78d0e4bc

I guess I should also mention that I'm using Vintageous. I dunno if that could cause the issue. I can try disabling some plugins if you think it might be some sort of plugin conflict.

zhengbli:
@johnfn The issue is related to #187, which appears to be a sublime text issue. Do you still have problems using the errorlist with the Vintageous plugin disabled?

typescript/libs/node_client.py
```diff
@@ -18,19 +18,34 @@
...
+    def sendCmdToWorker(self, cmd, cb): pass
+
     def sendCmdSync(self, cmd): pass
 
+    def sendCmdToWorkerSync(self, cmd, cb): pass
+    
     def sendCmdAsync(self, cmd, cb): pass
 
+    def sendCmdToWorkerAsync(self, cmd, cb): pass
+
```
hoanhtien: Refactor needed: Worker methods are mostly duplicate of server methods. Better to create worker and server as two instances of CommClient.

typescript/libs/view_helpers.py
```diff
@@ -57,9 +59,21 @@ def get_info(view):
...
+            # on the worker, which could be caused by starting the worker
+            # for the first time
+            if not IS_ST2:
+                if get_panel_manager().is_panel_active("errorlist"):
+                    info_on_worker = _file_map_on_worker.get(file_name)
+                    if not info_on_worker:
+                        _file_map_on_worker[file_name] = info
+                        open_file_on_worker(view)
+                        if view.is_dirty():
+                            if not view.is_loading():
```
hoanhtien: Combine into one if.

typescript/libs/view_helpers.py
```diff
@@ -213,10 +229,28 @@ def reload_buffer(view, client_info=None):
             client_info.change_count = info.modify_count
         client_info.pending_changes = False
 
+def reload_buffer_on_worker(view):
```
hoanhtien: There is a number of places we are calling reload_buffer but aren't calling reload_buffer_on_worker.

hoanhtien:
How can I trigger TypescriptGoToError command?

zhengbli:
By hitting enter on the error list item

hoanhtien:
Hitting enter doesn't work for me. Which part of the code is responsible for that? I am thinking it is `Default.sublime-keymap` but I do not find this file in the pull request. I may be missing something.

typescript/listeners/error_list.py
```diff
@@ -0,0 +1,130 @@
...
+    def handle_update_error_list_panel(self):
+        self.pending_update_error_list_panel -= 1
+        if self.pending_update_error_list_panel == 0:
+            self.update_error_list_panel()
+    
+    def update_error_list_panel(self):
+        log.debug("update error list panel")
+        output_lines = []
+        output_line_map = dict()
+        cur_line_number = 0
```
hoanhtien: This line can be removed.

typescript/listeners/error_list.py
```diff
@@ -0,0 +1,130 @@
...
+        log.debug(json_dict)
+        if json_dict["type"] != "event":
+            return
+
+        error_type = json_dict["event"]
+        if error_type not in ["syntaxDiag", "semanticDiag"]:
+            return
+
+        body = json_dict["body"]
+        if body is not None:
```
hoanhtien: Clearer if using:
```
if body is None:
   return
```

typescript/libs/view_helpers.py
zhengbli: The ``reload_buffer`` send reload commands to both the main server and the worker server (if up); and ``reload_buffer_on_worker`` only sends reload commands to the worker server. Therefore syncing shouldn't be a problem. And I cannot repro the problem you mentioned, the error message would go away as expected on the panel, though it may take a second.

typescript/libs/node_client.py
zhengbli: This is debatable. So I've tried to make two subclasses of CommClient, which leads to the problem that our ``ServerProxy`` only takes one ``CommClient`` instance at a time. However some methods like ``

typescript/listeners/error_list.py
zhengbli: I think it is clearer to have fewer outlets though

hoanhtien:
:+1:

PR #247, opened 2015-06-22
Update TS language service to release 1.5 version

### LINKED ISSUE #244: Wrong output path when using 'out' property in tsconfig
The `tsc` command line tool works perfectly, but when using the new build feature I am having an issue when using the `out` property to combine all of my TypeScript files into a single file.

To be clear, `outDir` is being ignored when `out` is used, which is the same behavior from the CLI tool. But I want to set my `out` path to `js/ts-bundle.js`. Works great with `tsc`, but with the Sublime Text plugin the `js/ts-bundle.js` is created inside the `src` directory.

My folder structure

```
Project
-- css
-- images
-- js
  -- lib
  -- src
index.html
tsconfig.json
```

I've tested various formats and setting the `sourceRoot` property, but nothing is working. Any help would be appreciated.

Here's my tsconfig.json file
```
{
	"compilerOptions": {
		"noImplicitAny": true,
		"removeComments": true,
		"out": "js/ts-bundle.js",
		"sourceMap": true
	},
	"files": [
		"js/src/fit-height.ts",
		"js/src/sidenav.ts",
		"js/src/external-links.ts",
		"typings/jquery/jquery.d.ts",
		"typings/tsd.d.ts"
	]
}
```

zhengbli:
Update tsserver.js, lib.d.ts and tsc.js to release 1.5 version. Also added the parts needed for the new build function.

Fix #244

### DIFF
--- a/tsserver/tsc.js
+++ b/tsserver/tsc.js
@@ -691,6 +691,36 @@ var ts;
         Ternary[Ternary["True"] = -1] = "True";
     })(ts.Ternary || (ts.Ternary = {}));
     var Ternary = ts.Ternary;
+    function createFileMap(getCanonicalFileName) {
+        var files = {};
+        return {
+            get: get,
+            set: set,
+            contains: contains,
+            remove: remove,
+            forEachValue: forEachValueInMap
+        };
+        function set(fileName, value) {
+            files[normalizeKey(fileName)] = value;
+        }
+        function get(fileName) {
+            return files[normalizeKey(fileName)];
+        }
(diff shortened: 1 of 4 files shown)


tsserver/tsc.js
```diff
@@ -691,6 +691,36 @@ var ts;
         Ternary[Ternary["True"] = -1] = "True";
```
mhegazy: I do not think we need to have another copy of the compiler.

mhegazy:
tsserver changes look good.

PR #248, opened 2015-06-22
Fix issue 245.

hoanhtien:
`view` can be `null` when using multiple column layout; add a null guard to avoid NoneType exception.


zhengbli:
:+1:

Issue #251, opened 2015-06-23
close_all should not terminate tsserver


hoanhtien:
Repro steps:
1. Open some ts files.
2. File > Close All Files, or Ctrl + Shift + P then "File: Close All".
3. Errors in console:
```
OSError: [Errno 22] Invalid argument
Traceback (most recent call last):
  File "C:\Program Files\Sublime Text 3\sublime_plugin.py", line 232, in on_close
    callback.on_close(v)
  File "C:\Users\tihoanh\AppData\Roaming\Sublime Text 3\Packages\TypeScript-Sublime-Plugin\typescript\listeners\listeners.py", line 229, in on_close
    cli.service.close(file_name)
  File "C:\Users\tihoanh\AppData\Roaming\Sublime Text 3\Packages\TypeScript-Sublime-Plugin\typescript\libs\service_proxy.py", line 113, in close
    self.__comm.postCmd(json_str)
  File "C:\Users\tihoanh\AppData\Roaming\Sublime Text 3\Packages\TypeScript-Sublime-Plugin\typescript\libs\node_client.py", line 174, in postCmd
    self.__serverProc.stdin.flush()
OSError: [Errno 22] Invalid argument
```
This is because tsserver terminates before all `on_close` callbacks are done.
4. Now open a ts file, you should see the typescript plugin does not work anymore.

jmolero:
The error highlighting still doesn't work after closing the project.

hoanhtien:
@jmolero Can you help elaborate the problem you are having?

jmolero:
Now when I close and reopen a project all the features like signature popup, quick reference, etc, work. None of them was working when reopening a project. The only feature that seems to stop working after closing and reopening the project is the error highlighting. I tried typing a random text and isn't marked in red. It only works if I close and reopen sublime text.

PR #253, opened 2015-06-23
No longer terminate tsserver upon close_all, close_window, and close_project command.

hoanhtien:
This fixes issue https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/251.
Remove `cli.service.exit()` from `close_all`, `close_window`, and `close_project` command handler.
Ignore expensive `on_activate` callbacks during `close_all` and `close_project` process.


typescript/listeners/listeners.py
```diff
@@ -227,7 +238,31 @@ def on_close(self, view):
             #         most_recent_used_file_list.remove(info)
             # notify the server that the file is closed
             cli.service.close(file_name)
+        if TypeScriptEventListener.__close_all_is_happening():
+            TypeScriptEventListener.__remove_close_all_view(view)
```
zhengbli: This can be simplified. You just need to check if the current number of opened views = 1, which eliminates the need to maintain a list of opened view

hoanhtien: Makes sense. Let me fix that.

typescript/listeners/listeners.py
```diff
@@ -227,6 +230,11 @@ def on_close(self, view):
             #         most_recent_used_file_list.remove(info)
             # notify the server that the file is closed
             cli.service.close(file_name)
+        if TypeScriptEventListener.about_to_close_all:
```
DanielRosenwasser: Add a newline above

typescript/listeners/listeners.py
```diff
@@ -227,6 +230,11 @@ def on_close(self, view):
...
+            window = sublime.active_window()
+            if window is None or not window.views():
```
DanielRosenwasser: newline above

typescript/listeners/listeners.py
```diff
@@ -10,11 +10,15 @@
 from .rename import RenameEventListener
 from .tooltip import TooltipEventListener
 
-
 class TypeScriptEventListener(sublime_plugin.EventListener):
     """To avoid duplicated behavior among event listeners"""
+
+    about_to_close_all = False
```
DanielRosenwasser: Leave a comment above about what this indicates of Sublime and why we need this

typescript/listeners/listeners.py
DanielRosenwasser: Leave a comment immediately above explaining what you are trying to avoid/accomplish

Issue #44, opened 2015-02-25
An error at the EOF has no span


JsonFreeman:
```ts
var map:
```

This gives the correct error ("Type expected"), but has no span if it is at the end of the file. In cases like these, VS uses the last character of the file as the span. That way the user sees that there is an error.

PR #258, opened 2015-06-25
Show error span at the last non-whitespace character of the file if the error is at EOF.

hoanhtien:
This fixes https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/44.

### DIFF
--- a/typescript/libs/view_helpers.py
+++ b/typescript/libs/view_helpers.py
@@ -316,3 +316,19 @@ def change_count(view):
             return info.modify_count
         else:
             return view.change_count()
+
+def last_non_whitespace_position(view):
+    """
+    Returns the position of the last non-whitespace character of <view>.
+    Returns -1 if <view> contains no non-whitespace character.
+    """
+    pos = view.size() - 1
+    while pos >= 0 and view.substr(pos).isspace():
+        pos -= 1
+    return pos
+
+def last_character_region(view):
+    """Returns the last non whitespace character under <sublime.Region> format"""
+    pos = last_non_whitespace_position(view)
+    return sublime.Region(pos, pos + 1)
+
--- a/typescript/libs/view_helpers.py
+++ b/typescript/libs/view_helpers.py
@@ -320,15 +320,15 @@ def change_count(view):
 def last_non_whitespace_position(view):
     """
     Returns the position of the last non-whitespace character of <view>.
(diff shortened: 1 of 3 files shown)


typescript/libs/view_helpers.py
```diff
@@ -315,4 +315,20 @@ def change_count(view):
...
             return info.modify_count
         else:
-            return view.change_count()
\ No newline at end of file
+            return view.change_count()
+
+def last_non_whitespace_position(view):
+    """
+    Returns the position of the last non-whitespace character of <view>.
+    Returns -1 if <view> contains no non-whitespace character.
```
DanielRosenwasser: only contains whitespace characters

typescript/libs/view_helpers.py
```diff
@@ -315,4 +315,20 @@ def change_count(view):
...
+    """
+    pos = view.size() - 1
+    while pos >= 0 and view.substr(pos).isspace():
+        pos -= 1
+    return pos
+
+def last_character_region(view):
+    """Returns the last non whitespace character under <sublime.Region> format"""
```
DanielRosenwasser: Returns a <sublime.Region> for the last non-whitespace character

typescript/listeners/idle.py
```diff
@@ -89,16 +89,22 @@ def show_errors(self, diagno_event_body, syntactic):
                         start = view.text_point(line, offset)
                         end = view.text_point(end_line, end_offset)
                         if end <= view.size():
-                            region = sublime.Region(start, end)
+                            if start == view.size() and start == end:
```
DanielRosenwasser: Leave a comment explaining why you need to do this

typescript/libs/view_helpers.py
```diff
@@ -315,4 +315,20 @@ def change_count(view):
...
\ No newline at end of file
+            return view.change_count()
+
+def last_non_whitespace_position(view):
+    """
+    Returns the position of the last non-whitespace character of <view>.
+    Returns -1 if <view> only contains non-whitespace characters.
+    """
+    pos = view.size() - 1
+    while pos >= 0 and view.substr(pos).isspace():
```
zhengbli: ``is_space()``

zhengbli:
:+1:

PR #259, opened 2015-06-25
Avoid running format_on_key for undo commands

### LINKED ISSUE #257: Auto-formatting in version 1.5 breaks undo history
The "Typescript Format On Key" action prevents undoing past the point at which the action was applied.

To replicate:

1. Ensure that you have the latest version of the TypeScript-Sublime-Plugin installed in Sublime Text
2. Create a new file and save as a file with a `.ts` extension so that the syntax gets set to TypeScript
3. Enter the following code: `function log(entry: {message: string`
4. Type `}`
5. Press Command + Z to undo
6. Nothing happens, because the action is re-applied as soon as it is undone.

I am running Sublime Text 3, Stable Channel, Build 3083.

zhengbli:
Fix #257

### DIFF
--- a/typescript/listeners/format.py
+++ b/typescript/listeners/format.py
@@ -27,7 +27,11 @@ class FormatEventListener:
             redo_command = view.command_history(1)[0]
             log.debug("last_command:{0}, args:{1}".format(last_command, args))
             log.debug("redo_command:{0}".format(redo_command))
-            if last_command == "insert" and redo_command != "typescript_format_on_key":
+            if redo_command != "":
+                # in an undo session, avoid running format_on_key
+                return 
+
+            if last_command == "insert":
                 pos = info.prev_sel[0].begin()
                 if ";" in args["characters"]:
                     view.run_command("typescript_format_on_key", {"key": ";"})
--- a/typescript/listeners/format.py
+++ b/typescript/listeners/format.py
@@ -1,4 +1,4 @@
-from .event_hub import EventHub
+﻿from .event_hub import EventHub
 from ..libs.view_helpers import *
 from ..libs.logger import log
 from ..libs import cli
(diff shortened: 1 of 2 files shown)


typescript/listeners/format.py
```diff
@@ -24,8 +24,10 @@ def on_modified_with_info(self, view, info):
             info.prev_sel[0].empty()
         ):
             last_command, args, repeat_times = view.command_history(0)
+            redo_command = view.command_history(1)[0]
             log.debug("last_command:{0}, args:{1}".format(last_command, args))
-            if last_command == "insert":
+            log.debug("redo_command:{0}".format(redo_command))
+            if last_command == "insert" and redo_command != "typescript_format_on_key":
```
DanielRosenwasser: We may want to rethink this approach - if there are any items to redo, then that means that the user just performed an undo, right? If there's anything that can be redone, then the user is traversing between undos and redos, and we should just hold back on doing anything.

Though I'm realizing, if a user redoes all of their operations, there is nothing on the redo list - but we wouldn't want to immediately format. Let's discuss this a bit.

__________________

If we do decide to go with the existing code,  I'd prefer you defined `redo_command` and performed the check inside this if, and put a comment before it like

```Python
# In this case, the user has requested we undo a formatting operation.
# In this case, 
redo_command = view.command_history(1)[0]
if redo_command == "typescript_format_on_key":
    return
```

zhengbli: The original assumption was that every insert command is followed by a ``format_on_key`` if enabled. Though on a second thought, checking what the redo command is seems to be redundant.

zhengbli: And if the user has redone every operation, the last one cannot be ``insert`` when the ``format_on_key`` was enabled.

hoanhtien:
Doesn't look like it is working. I try: `var x=1;` then undo.

hoanhtien:
Actually it works now after a few restarting. So :+1:.

Issue #264, opened 2015-06-29
Typescript plugin breaks the "auto_complete_commit_on_tab" option


ggrks:
```
// By default, auto complete will commit the current completion on enter.
// This setting can be used to make it complete on tab instead.
// Completing on tab is generally a superior option, as it removes
// ambiguity between committing the completion and inserting a newline.
    "auto_complete_commit_on_tab": true
```
This option does not work when typescript plugin is enabled.

hoanhtien:
@ggrks A quick fix is to remove this part from `Default.sublime-keymap`:
```
{
    "keys": [ "enter" ],
    "command": "commit_completion",
    "context": [
        { "key": "auto_complete_visible", "operator": "equal", "operand": true }
    ]
}
```

ggrks:
@hoanhtien 
It works, thanks.

PR #266, opened 2015-06-30
Remove key binding: enter -> commit_completion.

hoanhtien:
Two reasons to remove this:

1. This is redundant because enter triggers `commit_completion` command by default.
2. The `"auto_complete_commit_on_tab": true` setting won't work if this key binding presents. Issue https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/264.


zhengbli:
:+1:

Issue #262, opened 2015-06-26
Something wrong with TypeScript syntax highlights


NekR:
This is how it looks now, once (recently) I made upgrade to latest version in Package Control.

![ts](https://cloud.githubusercontent.com/assets/1189265/8380243/5cdd1a58-1c2d-11e5-939d-5f7001b69b3b.png)
![ts2](https://cloud.githubusercontent.com/assets/1189265/8380244/5cdd86c8-1c2d-11e5-9cef-f900015e04b1.png)
![ts3](https://cloud.githubusercontent.com/assets/1189265/8380245/5ce02f54-1c2d-11e5-813b-322b002c44f7.png)

zhengbli:
Is the ``getLevelTiles`` rendered correctly in your theme? If so, can you show us what is the content of the function? I know template strings (backtick) may cause issues here

hoanhtien:
Confirmed this repros when installing via Package Control.

NekR:
> Is the getLevelTiles rendered correctly in your theme?

Nope. It seems that problem is in ```for``` loops, because problem always occurs after them. Here is 2 screenshots, one with my current theme and one with popular Monokai theme:
![ts](https://cloud.githubusercontent.com/assets/1189265/8387271/e4e8a160-1c5f-11e5-802d-90d3542b9538.png)
![ts2](https://cloud.githubusercontent.com/assets/1189265/8387272/e4eb3c2c-1c5f-11e5-9822-adefa57fdcf4.png)

hoanhtien:
@NekR We just released v0.1.6 that includes the fix for this bug. Can you give it a try?

NekR:
@hoanhtien of course :-)

Just tried, all seems good at a first glance :+1: Thanks for the update!

PR #267, opened 2015-06-30
Fix the syntax highlighting issues with "<" sign

zhengbli:
Fix issue #262

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -133,7 +133,7 @@ repository:
       '2': { name: keyword.operator.ts }
       '3': { name: entity.name.function.ts }
       '4': { name: keyword.operator.ts }
-    end: '(?=\}|;|,)|(?<=\})'
+    end: '(?=\}|;|,|\n)|(?<=\})'
     patterns:
     - include: '#comment'
     - include: '#type-parameters'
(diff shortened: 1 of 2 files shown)


TypeScript.YAML-tmLanguage
```diff
@@ -176,7 +176,7 @@ repository:
     beginCaptures:
       '1': { name: storage.type.function.ts }
       '2': { name: entity.name.function.ts }
-    end: (?=;|\})|(?<=\})
+    end: (?=;|\}|\n)|(?<=\})
```
hoanhtien: As discussed, this may not work for:
```javascript
function
foo() {
}
```
But since this is rarely used, I think it should be fine.

TypeScript.YAML-tmLanguage
```diff
@@ -361,6 +360,7 @@ repository:
     - include: '#logic-operator'
     - include: '#assignment-operator'
     - include: '#storage-keyword'
+    - include: '#function-call'
```
hoanhtien: Personally I don't see `#function-call` being used anywhere. Cases that I have tried:
```
foo(1);
var x = foo(1);
let x = foo<number>(1, 2);
```
Considering remove this `#function-call` if it is not necessary. Or perhaps there is a problem somewhere else.

hoanhtien:
:+1:

Issue #277, opened 2015-07-07
Finishing a line with semicolon deletes letters from the next line


qbolec:
I try hard to use 2 spaces indentation everywhere.
```
    "tab_size": 2,
    "translate_tabs_to_spaces": true,
    "use_tab_stops": true,
```
TypeScript plugin fights against it, and while this is a hussle, I am ok with that.
But recently I observed that from time to time (maybe it's a timing issue?) several letters from next line are eaten.
In the attached screenshot I've tried to type in a line between lines 127 and 129.
I put my cursor at the end of line 127, hit enter, typed in by hand `this.trigger('active');`.
As soon as I typed `;` at the end of line 128, the line 129 got reformated and mangled (you can see that original `this.debouncedInactiveTrigger()` become shifted four spaces to the right and that word `this` become `th`.
What's strange is that I can reproduce this almost every time, but the exact prefix of line 128 which is affected is sometimes different. Sometimes it becomes `thidebouncedInactiveTrigger();` for instance.
![tsbug](https://cloud.githubusercontent.com/assets/133093/8542653/91944132-2497-11e5-9a8c-ef3ba22e93f9.png)

I should add, that I also have T3S plugin installed, but I've disabled it, and the problem persists.

hoanhtien:
@qbolec I am not able to repro this issue. Would you mind sending us a stripped-down version of your project?

qbolec:
Well, if I strip it down, the problem manifests slightly different:/
For example if I have a folder with just a single file app.ts 
```
class A{
  alice(){
    this.charlie();
    this.charlie();
  }
  bob(){

  }
  charlie(){

  }
}
```
(with linux line endings and no trailing spaces) and try to add `this.bob();` between the two `this.charlie();` the only strange effect I observe is that the second `this.charlie();` gets indented by additional 2 spaces. The strangeness is in the non-determinism of this - sometimes the indentation does not change, sometimes it does. I've noticed that I have more luck when adding the line for the first time to a freshly opened unmodified file.

hoanhtien:
@qbolec Thank you for the repro. This should be resolved by adding the following setting: `"indent_size":2` and restarting Sublime Text. Can you help check if this setting resolves your original issue (`this.debouncedInactiveTrigger();` gets wrongly formatted) too?

qbolec:
It did not solve the problem neither in the original app nor in the repro :/
[...]

PR #282, opened 2015-07-08
Reload buffer after insert_snippet (for Sublime 2).

hoanhtien:
This fixes https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/277.

We reload buffer in `on_post_text_command` callback. However Sublime 2 does not have this `on_post_text_command` API, so we implement this logic for Sublime 2 in `on_modified`.


zhengbli:
:+1:

PR #283, opened 2015-07-08
Add syntax highlighting for .tsx file.

hoanhtien:
Screenshot:
image


zhengbli:
@RyanCavanaugh could you take a look?

TypeScriptReact.YAML-tmLanguage
```diff
@@ -0,0 +1,720 @@
+# Similar to TypeScript.YAML-tmLanguage except:
+#   - Remove type assertion (cast.expr.ts)
```
billti: If this is largely a copy, did you consider having the source.tsx language just include the source.ts language? See https://www.sublimetext.com/forum/viewtopic.php?f=6&t=16560#p62430

TypeScriptReact.YAML-tmLanguage
```diff
@@ -0,0 +1,720 @@
...
+  object-member:
+    name: meta.object.member.tsx
+    begin: '[a-zA-Z_$][\w$]*\s*:'
+    end: (?=,|\})
+    patterns:
+    - include: '#expression'
+
+  expression-operator:
+    name: keyword.operator.tsx
+    match: =>|\b(delete|export|import|in|instanceof|module|namespace|new|typeof|void)\b
```
RyanCavanaugh: `as` should be here, not with the control flow keywords

TypeScriptReact.YAML-tmLanguage
hoanhtien: Haven't found a good way to do this. I can move independent scopes like #string, #keyword, #regex, #comment, #operator to a shared common syntax file. This can save ~100 lines each file. I cannot do the same thing with scopes that involves #expression because of its recursive structure.

billti: OK. Thanks for investigating. Maybe just create 2 different syntax files then, and try to keep the differences grouped together so they can be easily compared for any divergence as we iterate. Thanks!

RyanCavanaugh:
Seems good. Once this is merged in I'll try it in my sample projects and see how it looks

Issue #286, opened 2015-07-10
FormatBlock command does not work


hoanhtien:
Place the cursor inside this block and trigger FormatBlock command (`Ctrl + Shift + ]`)
```
{
var x=1;
var p= "";
}
```
**Expected**
```
{
	var x = 1;
	var p =  "";
}
```
**Actual result**
The code block does not get formatted. The following error shows in the console:
```
Traceback (most recent call last):
  File "C:\Program Files\Sublime Text 3\sublime_plugin.py", line 550, in run_
    return self.run(edit, **args)
TypeError: run() got an unexpected keyword argument 'insertKey'
```

PR #287, opened 2015-07-10
Parameter passed to typescript_format_on_key should be insert_key instead of insertKey

hoanhtien:
This fixes https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/286.


billti:
:+1:

Issue #288, opened 2015-07-10
Coloring screwed up after a name that has "function" in it


roblourens:
![capture](https://cloud.githubusercontent.com/assets/323878/8628858/279da838-270b-11e5-9ca2-5a84e4d8afd7.PNG)

PR #289, opened 2015-07-13
"Function" keyword should be followed by \b.

hoanhtien:
Make sure what comes after `function` is not a word character. 
This fixes https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/288.


billti:
:+1:

zhengbli:
:+1:

Issue #292, opened 2015-07-15
Incorrect coloring using generics


dbaeumer:
See screen shot
![capture](https://cloud.githubusercontent.com/assets/1931590/8696808/003d65e2-2af0-11e5-99fa-ea4f4798de5f.png)

PR #294, opened 2015-07-16
Right side of arrow function should also end at >

hoanhtien:
This fixes https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/292.

Before:
image

After:
image


zhengbli:
:+1:

Issue #293, opened 2015-07-15
Exception when drag-and-drop a typescript tab


hoanhtien:
![draganddropexception](https://cloud.githubusercontent.com/assets/1707813/8709948/09d60cde-2afb-11e5-829e-c99db3ff2eda.gif)

Not critical but we don't like exceptions anyway.

PR #297, opened 2015-07-17
Add a null check after for .active_view().

hoanhtien:
This fixes https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/293.


zhengbli:
:+1:

PR #299, opened 2015-07-20
Update ts to1.5 (with support for errorlist and project info)

zhengbli:


tsserver/tsc.js
```diff
@@ -31087,4 +31087,4 @@ var ts;
     }
 })(ts || (ts = {}));
 ts.executeCommandLine(ts.sys.args);
-//# sourceMappingURL=file:///C:/Users/Zhengbo/Documents/GitHub/TypeScript/built/local/tsc.js.map
```
hoanhtien: The `tsc.js` file in #243 has one more change. It is important?

zhengbli: Updated #243. It is important fix for issue #244, which is already in master and cherry-picked here.

hoanhtien:
:+1:

Issue #296, opened 2015-07-16
Notify user of TypeScript version mismatch


hoanhtien:
After the plug-in is upgraded automatically by PackageControl, we should notify user to upgrade his/her global TypeScript compiler and server (the ones installed with npm) if he/she has not done so.

PR #300, opened 2015-07-21
Notify user to upgrade npm typescript.

hoanhtien:
- After the plugin is upgraded automatically by PackageControl, we show a message to notify user to upgrade his/her global TypeScript compiler and server (the ones installed with npm). 
- We don't show the message if there is no npm or no global typescript installed.
- If we show the message, we only do it once (until the plugin is updated again).

https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/296.

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,5 +1,6 @@
 import sys
 import os
+import subprocess
 
 if sys.version_info < (3, 0):
     from typescript.libs import *
@@ -83,6 +84,7 @@ def plugin_loaded():
     else:
         print("ref view not found")
     log.debug("plugin_loaded ended")
+    _check_typescript_version()
 
 
 def plugin_unloaded():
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -96,3 +98,56 @@ def plugin_unloaded():
         if ref_info:
             ref_view.settings().set('refinfo', ref_info.as_value())
     cli.service.exit()
+
+
+_UPDATE_NPM_MESSAGE = "Warning from TypeScript Plugin: We detect a global \
```
DanielRosenwasser: Use a multiline string

```Python
"""Warning from TypeScript 
blah
blah
blah
"""
```

main.py
```diff
@@ -96,3 +98,56 @@ def plugin_unloaded():
...
+npm typescript module and it is outdated. Consider upgrading this module by \
+running \"npm install -g typescipt\". TypeScript Plugin still works \
+regardless."
```
DanielRosenwasser: I'd prefer

```
UPDATE_NPM_MESSAGE = "A newer version of TypeScript is available. Consider upgrading by running 'npm install -g typescript'."
```

main.py
```diff
@@ -96,3 +98,56 @@ def plugin_unloaded():
...
+
+            if npm_tsc_version < plugin_tsc_version:
+                sublime.message_dialog(_UPDATE_NPM_MESSAGE)
+
+                # Update the version in setting file so we don't show this
+                # message twice.
+                settings.set("typescript_plugin_tsc_version", plugin_tsc_version)
+                sublime.save_settings("Preferences.sublime-settings")
+
+    except Exception as error:
```
DanielRosenwasser: Is there any specific part we anticipate failure for?

main.py
```diff
@@ -96,3 +98,56 @@ def plugin_unloaded():
...
+    return _execute_cmd_and_parse_version_from_output(cmd)
+
+def _execute_cmd_and_parse_version_from_output(cmd):
+    output = subprocess.check_output(cmd, shell=True).decode('UTF-8')
+
+    # Use regex to parse the verion number from <output> e.g. parse
+    # "1.5.0-beta" from "message TS6029: Version 1.5.0-beta\r\n".
+    match_object = re.search("Version\s*([\w.-]+)", output, re.IGNORECASE)
+    if match_object is None:
+        raise Exception("Cannot parse version number from ouput: \"{0}\"".format(output))
```
DanielRosenwasser: Just use single quotes

main.py
hoanhtien: `subprocess.check_output()` throws exception if `node` or `tsc` is not in the PATH.

main.py
```diff
@@ -96,3 +98,56 @@ def plugin_unloaded():
...
+
+        if cached_plugin_version != plugin_tsc_version:
+            # The version number getting from the tsc command is different to
+            # the version number stored in the setting file. This means user
+            # has just updated the plugin.
+
+            npm_tsc_version = _get_npm_tsc_version()
+
+            if npm_tsc_version < plugin_tsc_version:
+                sublime.message_dialog(_UPDATE_NPM_MESSAGE)
```
billti: It doesn't necessarily have to be that the plugin version is newer. It may be they are running a newer command-line compiler (e.g. they may have a nightly build or a beta). That could still lead to inconsistencies between the plug-in behavior and 'tsc' on the command line. Just check for 'not equal'

main.py
billti: In general, this is a courtesy check. We shouldn't crash the plug-in if it fails. I like leaving this swallowing in regardless.

main.py
billti: Per below comment about version, we should just say something like: "Detected command-line TypeScript compiler version '{0}'. The TypeScript Sublime Text plugin is using compiler version '{1}'. There may be differences in behavior between releases. 

To update your command-line TypeScript compiler to the latest release, run 'npm update -g typescript'."

billti:
A couple of minor comments, otherwise :+1:

Issue #279, opened 2015-07-07
popup_manager only works for builds 3070+


joneshf:
This should be mentioned somewhere, as otherwise it's nigh impossible to figure out why it doesn't work.

Original discussion on [gitter](https://gitter.im/Microsoft/TypeScript-Sublime-Plugin?at=55966195fda0bd877e081016).

zhengbli:
We did mention in the [readme](https://github.com/Microsoft/TypeScript-Sublime-Plugin#platform-support) that:
> the use of a Sublime Text 3 build >= 3070 is recommended, as this provides a popup API used for tool tips.

You should have intellisense (a.k.a auto completion) even if you are using an older version. Are you not getting that? Thanks.

joneshf:
Oh. I missed that.

Could that feature get its own section [here](https://github.com/Microsoft/TypeScript-Sublime-Plugin#features) that explains how to use it as well? I think this feature is great. I wasn't even aware sublime added a popup feature until I installed this package.

And yes, completion works. It's just that the popup_manager does not work for any build older than 3070 (including st2).

zhengbli:
Sure. The readme file is a bit out of date. Thanks for pointing out!

hoanhtien:
Readme updated. Closing this thread.

joneshf:
Awesome, thanks! I just have one suggestion. It'd be nice to see that information near the screenshot since the naming isn't uniform as to what it's called when searching text (intellisense, popup, completion, etc) so it might be hard to find out why it doesn't work when you need it to. Searching for the name was part of my problem originally.

I'll submit a PR with what I mean.

PR #306, opened 2015-07-23
Mention build version for signature popup

joneshf:
Re: https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/279#issuecomment-124083731

Again, the reason for this change is that it's kind of tough to know what to search the text for, but it's easier to see visually what feature the signature popup is. So if you're like me, and you come here wondering why the completion/intellisense/popup/tooltip/type info/etc isn't working, you don't have to spend 20 minutes trying to figure out what it was named.

Hopefully that reason makes sense. If the wording of the change should be different, let me know.


hoanhtien:
Thank you. This is useful.

Issue #275, opened 2015-07-06
Comment between type and initializer not colored correctly


dbaeumer:
var x: string /* comment */ = "foo";

zhengbli:
@hoanhtien does this get fixed in #238 ?

hoanhtien:
@zhengbli No, #238 only fixes comments in the `new` expression.

PR #307, opened 2015-07-23
Colorize string at certain syntactically incorrect positions.

hoanhtien:
This fixes https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/227.

Before:
image

After:
image

PR #308, opened 2015-07-23
Colorize string and comment between type and initializer.

hoanhtien:
This fixes https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/275.

Before:
image

After:
image

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -241,6 +241,8 @@ repository:
     end: (?=$|[,);\}\]]|//)|(?==[^>])|(?<=[\}>\]\)]|[a-zA-Z_$])\s*(?=\{)
     patterns:
     - include: '#type'
+    - include: '#string'
+    - include: '#comment'
 
   type:
     name: meta.type.ts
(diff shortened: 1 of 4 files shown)


TypeScript.YAML-tmLanguage
```diff
@@ -241,6 +241,8 @@ repository:
     end: (?=$|[,);\}\]]|//)|(?==[^>])|(?<=[\}>\]\)]|[a-zA-Z_$])\s*(?=\{)
     patterns:
     - include: '#type'
+    - include: '#string'
```
zhengbli: Why string?

hoanhtien: I was fixing "Colorize string at certain syntactically incorrect positions." https://github.com/Microsoft/TypeScript-Sublime-Plugin/pull/307 so I think people may want to have string here too.

zhengbli:
:+1:

TypeScript.YAML-tmLanguage
DanielRosenwasser: We're also thinking about having string literal types eventually too, so it will also make actual sense in the future. That also means that the following is going to be a thing as well:

```TypeScript
type State = "Ready" | "Running" | "Recalibrating" | "Reticulating Splines" | "Returning";
```

DanielRosenwasser:
:+1:

Issue #39, opened 2015-02-24
Support occurrences highlighting


DanielRosenwasser:
When the cursor resides on a word, it would be helpful to populate appropriate regions.

Looks like [`add_regions`](https://www.sublimetext.com/docs/3/api_reference.html#sublime.View) might be the relevant API, with `sublime.DRAW_NO_FILL` for now.

Consider the Roslyn timeouts:
- 1.5 seconds after edits
- 50 milliseconds for non-destructive events (i.e. arrow keys, clicks)

hoanhtien:
I'm thinking of implementing this feature. To find occurrences, are we planning to use `occurrences` command (looks costly) or we gonna keep it simple by using a regex match?

PR #316, opened 2015-07-30
Add highlight occurrences

hoanhtien:
+ Refactor idle.py
+ Add highlight occurrences features (https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/39). Enable by default. To disable, use setting `'typescript_highligh_occurrences': false`. This works on ST3 only.

image


DanielRosenwasser:
Actually, I believe we are deprecating occurrences in favor of document highlights. It hasn't been exposed in the server though. I don't remember what the specific differences are, but you can ask @CyrusNajmabadi. This could be a decent temporary solution, but eventually we should move over if getOccurrences is deprecated.

typescript/listeners/idle.py
```diff
@@ -3,34 +3,81 @@
...
+        """
+        If selection is idle (cursor is not moving around), update the status
+        line (error message or quick info, if any).
+        """
+        log.debug("on_selection_idle")
+        view = active_view()
+        info = get_info(view)
+        if info:
+            self.update_status(view, info)
+            if not IS_ST2 and view.settings().get('typescript_highlight_occurrences', 'true'):
```
DanielRosenwasser: Is this going to be on by default?

typescript/listeners/idle.py
```diff
@@ -179,8 +181,43 @@ def update_status(self, view, info):
...
+            start_point = view.text_point(start_line, start_offset)
+
+            end_line, end_offset = extract_line_offset(occurrence['end'])
+            end_point = view.text_point(end_line, end_offset)
+
+            occurrence_regions.append(sublime.Region(start_point, end_point))
+
+        view.add_regions(region_key, occurrence_regions, 'comment', '',
+                         sublime.DRAW_NO_FILL |
+                         sublime.DRAW_NO_OUTLINE |
```
DanielRosenwasser: Can we see what it looks like with an outline instead of underlining?

typescript/listeners/idle.py
hoanhtien: Yes it is.

hoanhtien:
@DanielRosenwasser @mhegazy Maybe I can help with exposing document highlights in the server.

hoanhtien:
Updated this pull request to use document highlights (https://github.com/Microsoft/TypeScript/pull/4139).

typescript/listeners/idle.py
```diff
@@ -3,34 +3,81 @@
...
+        log.debug("on_idle")
+        view = active_view()
+        info = get_info(view)
+        if info:
+            self.request_errors(view, info, 500)
+
+    def on_selection_idle(self):
+        """
+        If selection is idle (cursor is not moving around), update the status
+        line (error message or quick info, if any).
```
zhengbli: also for document hightlight

typescript/listeners/idle.py
```diff
@@ -179,8 +181,46 @@ def update_status(self, view, info):
...
+        region_key = 'occurrences'
+        view.erase_regions(region_key)
+
+        if not response['success']:
+            return
+
+        occurrence_regions = []
+
+        for file_highlight in response['body']:
+            if file_highlight['file'] != view.file_name().replace('\\', '/'):
```
zhengbli: If I remember correctly, the tsserver changes the file name t olower case if the OS' file system is case-insensitive. Would that cause problems here? for example on or linux. Also, as we are comparing filenames in multiple places, it might worth it to extract that part to a individual function.

typescript/listeners/idle.py
```diff
@@ -179,8 +181,46 @@ def update_status(self, view, info):
...
+                continue
+
+            for occurrence in file_highlight['highlightSpans']:
+                start_line, start_offset = extract_line_offset(occurrence['start'])
+                start_point = view.text_point(start_line, start_offset)
+
+                start_line, start_offset = extract_line_offset(occurrence['start'])
+                start_point = view.text_point(start_line, start_offset)
```
zhengbli: Why do you do the extraction twice?

zhengbli:
Apart from several details that I left comments on, the implementation looks good to me.

typescript/listeners/idle.py
hoanhtien: I tested on Window and Linux and found no problem.

PR #318, opened 2015-08-01
Make command name formatting consistent with others

danquirk:
This:
image
instead of:
image


hoanhtien:
:+1:

DanielRosenwasser:
I have a question: why are we using PascalCase for these commands? These are user-facing actions.

hoanhtien:
"Navigate To Symbol" looks easier-to-read than "NavigateToSymbol". @danquirk Could you change to this format and merge?

PR #325, opened 2015-08-14
Prevent unnecessary calls to service.configure

### LINKED ISSUE #187: Settings change event triggering unnecesary calls to reconfig_file

jgoz:
The [Sublime API is somewhat unclear](https://www.sublimetext.com/docs/3/api_reference.html#sublime.Settings), but the `on_change` argument for `Settings.add_on_change` will be called whenever *any* setting changes. The `key` argument is simply a way of referring to the callback in a subsequent call to `Settings.clear_on_change`. This plugin is incorrectly assuming that `key` is the setting name will trigger the supplied callback.

Two changes are included here:
- Only register on_change handler once, since it will be triggered on every setting change
- Check previous format options against current values to detect a change; only call `configure` if one of the values has actually changed

I chose to store the previous values in the view settings because that seems to be the only persistent store for a view. If there is a better way to do this, please let me know.

Fixes #187.

### DIFF
--- a/typescript/libs/view_helpers.py
+++ b/typescript/libs/view_helpers.py
@@ -156,30 +156,37 @@ def reconfig_file(view):
     view_settings = view.settings()
     tab_size = view_settings.get('tab_size', 4)
     indent_size = view_settings.get('indent_size', tab_size)
-    translate_tab_to_spaces = view_settings.get('translate_tabs_to_spaces', True)
-    format_options = {
-        "tabSize": tab_size,
-        "indentSize": indent_size,
-        "convertTabsToSpaces": translate_tab_to_spaces
-    }
-    cli.service.configure(host_info, view.file_name(), format_options)
+    translate_tabs_to_spaces = view_settings.get('translate_tabs_to_spaces', True)
+
+    prev_format_options = view_settings.get('typescript_plugin_format_options')
+    if prev_format_options == None \
+        or prev_format_options['tabSize'] != tab_size \
+        or prev_format_options['indentSize'] != indent_size \
+        or prev_format_options['convertTabsToSpaces'] != translate_tabs_to_spaces:
+        format_options = {
+            "tabSize": tab_size,
+            "indentSize": indent_size,
+            "convertTabsToSpaces": translate_tabs_to_spaces
+        }
+        view_settings.set('typescript_plugin_format_options', format_options)
+        cli.service.configure(host_info, view.file_name(), format_options)
+        return True
+    return False
 
 
 def set_file_prefs(view):
     settings = view.settings()
     settings.set('use_tab_stops', False)
-    settings.add_on_change('tab_size', tab_size_changed)
-    settings.add_on_change('indent_size', tab_size_changed)
-    settings.add_on_change('translate_tabs_to_spaces', tab_size_changed)
+    settings.add_on_change('typescript_plugin_settings_changed', settings_changed)
 
 
-def tab_size_changed():
+def settings_changed():
     view = active_view()
     if view is None:
         return
-    reconfig_file(view)
-    client_info = cli.get_or_add_file(view.file_name())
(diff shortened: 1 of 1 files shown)


typescript/libs/view_helpers.py
```diff
@@ -156,30 +156,37 @@ def reconfig_file(view):
...
 
-def tab_size_changed():
+def settings_changed():
     view = active_view()
     if view is None:
         return
-    reconfig_file(view)
-    client_info = cli.get_or_add_file(view.file_name())
-    client_info.pending_changes = True
+    if reconfig_file(view):
```
DanielRosenwasser: Can you add a doc comment specifying what `reconfig_file` returns?

jgoz: Done.

hoanhtien:
:+1:

PR #329, opened 2015-08-21
Add is_open flag to FileInfo class to resend open command when reopen a file

zhengbli:
Currently ST only sends an "OpenFile" command to the server when the file is opened for the first time ever. If a file has been opened before, no "OpenFile" command would be sent.

This became a problem when TS 1.6 added a tsconfig file watcher to pick up changes in tsconfig file on the fly. The project that a file belongs to may have changed between the time it was closed last time and the time it was reopened. The trigger for updating project structure is placed in the callback function of the "OpenFile" command. Consequently, that wouldn't work here.

This PR fixes the situation by adding a "is_open" flag for the FileInfo corresponding to each view opened.


zhengbli:
@billti could you take a look?

typescript/libs/view_helpers.py
```diff
@@ -36,7 +37,7 @@ def __init__(self, filename, cc):
 _file_map_on_worker = dict()
 
 
-def get_info(view):
+def get_info(view, open_if_not_found=True):
```
billti: This param name is confusing. What does it mean exactly? (How do you open something if you cant find it?)

typescript/libs/view_helpers.py
```diff
@@ -46,33 +47,35 @@ def get_info(view):
...
-                info.client_info = cli.get_or_add_file(file_name)
-                set_file_prefs(view)
-                _file_map[file_name] = info
-                # Open the file on the server
-                open_file(view)
-                if view.is_dirty():
-                    if not view.is_loading():
-                        reload_buffer(view, info.client_info)
+            if open_if_not_found:
+                if not info or info.is_open is False:
```
billti: Should this block be inside the `open_if_not_found` condition? I would think the below still needs to execute `if not info`.

typescript/libs/view_helpers.py
billti: Ah, got it now. Maybe `open_if_not_cached` is a better name?

billti:
I'm trying to make sure I'm clear here: So does the server already contain logic to reload the project if it detects tsconfig.json has changed? Is this just to force Sublime to go reload any info for the file it already has open? I would have thoughts the server would detect and update the project based on changes anyway, and then the next time you reopen the view on the client it would re-request the errors, etc... for the file.

typescript/libs/view_helpers.py
zhengbli: I agree ``open_if_not_cached`` sounds more reasonable

zhengbli:
There are cases when the tsconfig.json would not change but the project structure changed. For example, a tsconfig file without a ``Files`` array implicitly includes every file in this folder; then a new file is added to this folder, the project should be updated accordingly, however the tsconfig file watcher wouldn't be triggered as this file is not changed. For now we update the project when the newly added file is opened by client.

Also, say the file was in a configured project, and was deleted from it after it is closed. The server would not rebuild an inferred project for a closed file. Therefore if we don't force the update, the file wouldn't have a project when reopened.

typescript/libs/view_helpers.py
```diff
@@ -13,6 +13,7 @@ def __init__(self, filename, cc):
         self.filename = filename
         # 'pre_change_sent' means the change to this file is already sent to the server
         # used between 'on_text_command' and 'on_modified'
```
hoanhtien: This comment should be put on top of line 17.

typescript/libs/view_helpers.py
hoanhtien: Suggestion: The following would make it easier to read:
```Python
49  info = _file_map.get(file_name) 
50  if not info or info.is_open is False:
51      if open_if_not_found:
```
or
```Python
49  info = _file_map.get(file_name) 
50  if (not info or info.is_open is False) and open_if_not_found:
```

zhengbli: The catch is that I want to avoid doing anything on the worker server either if the ``open_if_not_cached`` is False (line 69). The situation applies when I close a file and call ``get_info`` in the  ``on_activated`` function.

zhengbli:
@hoanhtien can you take a look after the updates

hoanhtien:
:+1:

Issue #335, opened 2015-08-28
Coloring broken when using .default


dbaeumer:
Test case:

class A1 {
    b() {
        let d;
        d.default;
    }
    
    private static c() {       
    }
}

Observe: private and static is not colored as a keyword.

PR #336, opened 2015-08-28
'Default' or 'case' preceded by dot (.) should not be captured as part of switch-case expression.

hoanhtien:
This fixes issue https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/335.

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -23,7 +23,7 @@ repository:
 
   control-statement:
     name: keyword.control.ts
-    match: \b(break|catch|continue|debugger|declare|do|else|finally|for|if|return|switch|throw|try|while|with|super|case|default)\b
+    match: (?<!\.)\b(break|catch|continue|debugger|declare|do|else|finally|for|if|return|switch|throw|try|while|with|super|case|default)\b
 
   switch-case: 
     name: case.expr.ts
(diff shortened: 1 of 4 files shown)


TypeScript.YAML-tmLanguage
```diff
@@ -23,14 +23,14 @@ repository:
 
   control-statement:
     name: keyword.control.ts
-    match: \b(break|catch|continue|debugger|declare|do|else|finally|for|if|return|switch|throw|try|while|with|super)\b
+    match: \b(break|catch|continue|debugger|declare|do|else|finally|for|if|return|switch|throw|try|while|with|super|case|default)\b
```
billti: I left a comment on the test-case also... but perhaps we should check keywords are not preceded by a "." . Reserved words are valid identifierNames, but not valid identifiers (i.e. "foo.return" is valid and "return" is not a keyword here, but "var return" is invalid as 'return' is a reserved work in this position).

hoanhtien: I originally thought we wanted to keep it highlighted as keywords as Sublime does that for JS files.

hoanhtien:
Updated the pull request to no longer colorize reserved words if they are being used as identifiers:
Before:
image
After:
image

billti:
Thanks!

PR #339, opened 2015-08-29
Snippets update.

kungfusheep:
Hi - I noticed the access modifier in the method snippet wasn't a field, so that has been rectified in this PR. I've also added a snippet for `namespace`. Thanks.


billti:
Thanks for the contribution @kungfusheep . Merged!

PR #344, opened 2015-09-01
Allow user to configure the font size of signature popup

hoanhtien:
Related to issue https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/219.


zhengbli:
:+1:

PR #349, opened 2015-09-04
Colorize abstract keyword

hoanhtien:
This fixes https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/347.

Or maybe we should use a different color?

image


DanielRosenwasser:
:+1:

PR #355, opened 2015-09-14
Distinguish the less than sign and type parameter bracket

### LINKED ISSUE #285: Syntax highlighting issue when there is no space preceding <
Related to https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/265.

Issue:
image

Correct:
image

zhengbli:
Fix #285


zhengbli:
Unhandled issues found:
1.
```
class test<T> {
    
}

var s: test<string>;

```
2. 
```
f6 = foo < f<string>();
```

hoanhtien:
Not sure if we should get this PR in. `i < f<string>()` looks like a common case to me and the current syntax can handle cases like that and more complex cases.

Issue #352, opened 2015-09-09
Code completion does not work when working with *.tsx files


pristas-peter:
When working with TypeScriptReact files, code completion does not work. When I switch to TypeScript file type, completion works.

Example:

/// <reference path="../typings/react/react.d.ts" />
import React = require('react');

React. <- nothing is autocompleted.

PR #356, opened 2015-09-15
Enable all snippets and custom commands to work with tsx source files

hoanhtien:
Related to https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/352.

### DIFF
--- a/Default (Linux).sublime-keymap
+++ b/Default (Linux).sublime-keymap
@@ -4,7 +4,7 @@
         "command": "typescript_paste_and_format",
         "context": [
             { "key": "setting.typescript_auto_format", "operator": "equal", "operand": true },
-            { "key": "selector", "operator": "equal", "operand": "source.ts" }
+            { "key": "selector", "operator": "equal", "operand": "source.ts, source.tsx" }
         ]
     }
 ]
--- a/TypeScriptReact.sublime-settings
+++ b/TypeScriptReact.sublime-settings
@@ -1,0 +1,4 @@
+{
+    "auto_complete_triggers" : [ {"selector": "source.tsx", "characters": "."} ],
+    "use_tab_stops": false
+}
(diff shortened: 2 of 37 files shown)


Default (Linux).sublime-keymap
```diff
@@ -4,7 +4,7 @@
         "command": "typescript_paste_and_format",
         "context": [
             { "key": "setting.typescript_auto_format", "operator": "equal", "operand": true },
-            { "key": "selector", "operator": "equal", "operand": "source.ts" }
+            { "key": "selector", "operator": "equal", "operand": "source.ts, source.tsx" }
```
billti: Interesting. I didn't realize you could just use a comma separated list here. That's handy! :smile:

TypeScriptReact.sublime-settings
```diff
@@ -0,0 +1,4 @@
+{
+    "auto_complete_triggers" : [ {"selector": "source.tsx", "characters": "."} ],
```
billti: Is there a reason this isn't just added to the existing sublime-settings file? (Not that it matters much).

billti:
:+1: Thanks!

TypeScriptReact.sublime-settings
hoanhtien: It doesn't work. I wasted some hours trying to make it work and eventually found the reason [here](http://sublime-text-unofficial-documentation.readthedocs.org/en/latest/customization/settings.html):

>file type settings need to carry the name of the .tmLanguage syntax definition for the file type. Thus, for the .py file type, whose syntax definition is contained in Python.tmLanguage, the corresponding settings files would be called Python.sublime-settings.

billti: Ah! That would be why I couldn't get it to work :confused: Thanks!

PR #359, opened 2015-09-18
Reduce queue timeout possibility.

hoanhtien:
+ Increase timeout for reading async response to 2 seconds.
+ In read_msg function, always run callback asynchronously to keep read_msg as small as possible.


zhengbli:
:+1:

Issue #362, opened 2015-09-22
Wrong highlighting when no trailing semicolon


antoinerousseau:
![capture d ecran 2015-09-22 15 29 35](https://cloud.githubusercontent.com/assets/160279/10019637/c4e00f7e-613e-11e5-9abb-25db15711461.png)

`private` is not highlighted if the first line does not end with a semicolon, even though this is correct code

hoanhtien:
@antoinerousseau Are `logger` and `placementsByPid` members of a class?

antoinerousseau:
Yep!

antoinerousseau:
look at the difference:
![capture d ecran 2015-09-24 11 55 51](https://cloud.githubusercontent.com/assets/160279/10070518/4d04b81c-62b3-11e5-8928-795d3c805cb9.png) ![capture d ecran 2015-09-24 11 56 03](https://cloud.githubusercontent.com/assets/160279/10070522/5076dc6e-62b3-11e5-8a6c-94704b46f6bc.png)

PR #365, opened 2015-09-25
Field-declaration should also end at \n.

hoanhtien:
Fixes https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/362.


zhengbli:
:+1:

Issue #358, opened 2015-09-17
'any' is not being colorized inside interface/class.


hoanhtien:
![image](https://cloud.githubusercontent.com/assets/1707813/9946104/1a5a40fc-5d46-11e5-9278-f42f61175407.png)

PR #366, opened 2015-09-25
Colorize `any` in class/interface.

hoanhtien:
Fixes https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/358.

### DIFF
--- a/TypeScript.YAML-tmLanguage
+++ b/TypeScript.YAML-tmLanguage
@@ -461,7 +461,7 @@ repository:
 
   storage-keyword:
     name: storage.type.ts
-    match: \b(number|boolean|string|var|let|function|const)\b
+    match: \b(number|boolean|string|any|var|let|function|const)\b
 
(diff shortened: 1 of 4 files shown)


TypeScript.YAML-tmLanguage
```diff
@@ -461,7 +461,7 @@ repository:
 
   storage-keyword:
     name: storage.type.ts
-    match: \b(number|boolean|string|var|let|function|const)\b
+    match: \b(number|boolean|string|any|var|let|function|const)\b
```
DanielRosenwasser: Does the JSX file need to be updated as well?

hoanhtien: Oh right, JSX does need to be updated.

hoanhtien: Updated.

zhengbli:
:+1:

Issue #367, opened 2015-09-29
Assigning to a member variable named interface makes syntax highlighting break


dbaeumer:
See picture
![bad highlighting](https://cloud.githubusercontent.com/assets/1931590/10178587/f4719b3c-6700-11e5-8f36-d8a7bbbcf13d.png)

hoanhtien:
Possibly related to https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/353.

Issue #353, opened 2015-09-11
Coloring broken when using `.class`


vilic:
![image](https://cloud.githubusercontent.com/assets/970430/9807866/c990004e-588b-11e5-8a61-37d4de79ec8f.png)

billti:
Tien, this fix might be similar to the other change recently made (i.e. ensure there is no "." before the keyword that starts the match on the class pattern `(?<!\.`..  )

PR #368, opened 2015-09-30
class/interface keyword should not be preceded with dot (.)

hoanhtien:
This fixes https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/367 and https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/353.


zhengbli:
:+1:

Issue #372, opened 2015-10-14
Syntax highlight for "class" and "interface" is broken


zhengbli:
![capture](https://cloud.githubusercontent.com/assets/1171301/10472111/8fe51564-71d1-11e5-889d-612727cd04a5.PNG)

PR #373, opened 2015-10-14
Fix colorization of class/interface inside namespace

hoanhtien:
Fixes https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/372.


zhengbli:
:+1:

PR #375, opened 2015-10-16
Add menu items for the plugin settings

zhengbli:
This PR exposes the plugin settings in the "Preferences" menu:
untitled

### DIFF
--- a/Main.sublime-menu
+++ b/Main.sublime-menu
@@ -18,10 +18,7 @@
             "caption": "TypeScript",
             "children": [
               {
-                "command": "open_file",
-                "args": {
-                  "file": "${packages}/TypeScript/Preferences.sublime-settings"
-                },
+                "command": "typescript_open_plugin_default_setting_file",
                 "caption": "Plugin Settings – Default"
               },
               {
@@ -35,10 +32,7 @@
                 "caption": "-"
               },
               {
-                "command": "open_file",
-                "args": {
-                  "file": "${packages}/TypeScript/TypeScript.sublime-settings"
-                },
+                "command":"typescript_open_ts_default_setting_file",
                 "caption": "TypeScript Settings – Default"
               },
               {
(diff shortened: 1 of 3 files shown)


Main.sublime-menu
```diff
@@ -1,9 +1,74 @@
...
+      {
+        "id": "package-settings",
+        "children": [
+          {
+            "caption": "TypeScript",
+            "children": [
+              {
+                "command": "open_file",
+                "args": {
+                  "file": "${packages}/TypeScript/Preferences.sublime-settings"
```
hoanhtien: For people cloning the plugin from github the folder name could be TypeScript-Sublime-Plugin?

zhengbli: You are right that is a problem in that case, though there seems to be no way around it.

typescript/commands/settings.py
```diff
@@ -0,0 +1,20 @@
+import sublime_plugin
+
+from ..libs.global_vars import *
+from ..libs import cli, logger
+import os
+
+class TypescriptOpenPluginDefaultSettingFile(sublime_plugin.WindowCommand):
```
DanielRosenwasser: Typescript -> TypeScript
Setting -> Settings

typescript/commands/__init__.py
```diff
@@ -59,5 +64,8 @@
     "TypescriptShowDoc",
     "TypescriptSignaturePanel",
     "TypescriptSignaturePopup",
-    "TypescriptBuildCommand"
+    "TypescriptBuildCommand",
+    "TypescriptOpenPluginDefaultSettingFile",
+    "TypescriptOpenTsDefaultSettingFile",
+    "TypescriptOpenTsreactDefaultSettingFile"
```
DanielRosenwasser: Actually, we can change the casing as part of a separate PR.

typescript/commands/settings.py
zhengbli: Though Sublime transforms this camel casing to snake casing, which means "TypeScript" will be transformed to "type_script" when referred to in sublime-menu files.

PR #388, opened 2015-11-02
Update tsserver with fixes

zhengbli:
Update the tsserver with the latest bit in release-1.7 branch, which contains changes made in https://github.com/Microsoft/TypeScript/pull/5408 that resolves crashes for OSX users.


billti:
Thanks! :+1:

Issue #235, opened 2015-06-15
ERROR: can not send request; node process not running


joltmode:
I installed TypeScript for Sublime for the first time and it worked like a charm. All of a sudden I am getting errors about node not running.

Excerpt from console:

    spawning node module: C:\Users\psycketom\Desktop\ST3\Data\Packages\TypeScript\tsserver\tsserver.js
    Found node executable at node
    2015-06-15 16:48:25,048: 12592: ERROR: can not send request; node process not running

I just edited my preferences, to point to the exact node path. No success either:

    spawning node module: C:\Users\psycketom\Desktop\ST3\Data\Packages\TypeScript\tsserver\tsserver.js
    Found node executable at C:\Program Files\nodejs\node.exe
    2015-06-15 16:58:35,129: 12448: ERROR: can not send request; node process not running

---

After numerous ST restarts, I found out that the problem is occasional. 

And environment info:
Windows 7 x64
node v0.12.3
Sublime Text 3, build 3083

---

What could be the problem?

DeezCashews:
Getting the same thing here.

Windows 7 x64
node v0.12.4
Sublime Text 3, build 3083

DanielRosenwasser:
How are you opening Sublime? Is node in your system's PATH variable?

zhengbli:
Is it happening on the latest bit on master? Looks like it can find the node path correctly, but cannot start it for some reason.

DeezCashews:
Yes node is on my path.  I am starting sublime from `D:\<somedir>\subl` as sublime is on my path too.  I can run `node D:\sublime3\Data\Packages\TypeScript\tsserver\tsserver.js` just fine from the command line also, but sublime3 was still giving the same errors in D:\sublime3\Data\Packages\TypeScript\TS.log.

zhengbli:
@DeezCashews  is the problem also occasional for you? And is your plugin installed from the package control or cloned from the repo?

DeezCashews:
Installed from the PackageControl.  I'm not sure if it's occasional or not right now.  I had it fixed temporarily I think by tweaking env, but then broke it on purpose to figure out solution for @joltmode but now I can't seem to get working again so I guess it could be occasional.

DeezCashews:
Ok, so I figured out WHAT is breaking it, but no idea why...

Executing D:\sublime3\subl.exe = errors above
Executing D:\sublime3\sublime_text.exe = lovely code assist from tsserver.js :)

[...]

PR #397, opened 2015-11-16
Clarify console messages.

zhengbli:
Related to #235. 
This PR makes the console clearer when the configured `node_path` is not a valid one, and also that a temped node path is not already really "found" successfully.

### DIFF
--- a/typescript/libs/node_client.py
+++ b/typescript/libs/node_client.py
@@ -257,7 +257,7 @@ class ServerClient(NodeCommClient):
             self.server_proc = None
         else:
             global_vars._node_path = node_path
-            print("Try spawning node executable from: " + node_path)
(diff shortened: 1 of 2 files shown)


typescript/libs/node_client.py
```diff
@@ -252,7 +257,7 @@ def __init__(self, script_path):
             self.server_proc = None
         else:
             global_vars._node_path = node_path
-            print("Found node executable at " + node_path)
+            print("Try spawning node executable from: " + node_path)
```
DanielRosenwasser: Remove "Try" (or change it to "Trying to spawn") - rigth now it suggests that the user needs to do something special.

DanielRosenwasser:
After nits are addressed :+1:

PR #398, opened 2015-11-19
Add notes about the new tmLanguage repo

zhengbli:


DanielRosenwasser:
:+1:

DanielRosenwasser:
We should also consider adding a comment to the YAML files like

```
NOTE: If you need to make an edit to this file, do so at
      https://github.com/Microsoft/TypeScript-TmLanguage
```

Issue #402, opened 2015-11-21
Disable analyzing and leaving syntax highlighting only


NekR:
How I can disable code analyzing but leaving syntax highlighting enabled? I have few quite big projects open at a same, most of the time I do not close tabs and sublime at all. But with TypeScript plugin it burns out my Macbook so its temperature is like I am playing modern game on high quality. Same on Windows and it takes 1GB+ of memory which isn't good when you have only 8 of it.

zhengbli:
Should be fixed by #406

Issue #164, opened 2015-04-28
Split language definitions from plugin


laktak:
Would it be possible to split the typescript language definition (tmLanguage, etc.) from the rest of the plugin? So it's basically what is was before you took over the `typescript` package?

I like the syntax highlighting but I prefer the Sublime way of code completion (where it allows me to complete anything instead of *intelli-sense*).

mhegazy:
We need an option to disable statement completion.

laktak:
Anything that disables all *background* operations would make me happy.

zhengbli:
Should be fixed by #406 
Now there is an `enable_typescript_language_service` option, by default it was set to `true`, the user can set it to `false` to turn off the language service.

PR #406, opened 2015-11-23
Add option to switch off the TypeScript language service

zhengbli:
Right now the TypeScript language service is always on. However there have been some request for providing the option to switch it off for reasons like faster browsering sessions on big projects etc.

Related: #402 #164

### DIFF
--- a/typescript/commands/signature.py
+++ b/typescript/commands/signature.py
@@ -60,7 +60,7 @@ class TypescriptSignaturePanel(TypeScriptBaseTextCommand):
 
 class TypescriptSignaturePopup(sublime_plugin.TextCommand):
     def is_enabled(self):
-        return TOOLTIP_SUPPORT and is_typescript(self.view)
+        return TOOLTIP_SUPPORT and is_typescript(self.view) and get_language_service_enabled()
 
     def run(self, edit, move=None):
         log.debug('In run for signature popup with move: {0}'.format(move if move else 'None'))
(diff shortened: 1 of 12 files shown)


typescript/commands/signature.py
```diff
@@ -60,7 +60,7 @@ def on_selected(self, index):
 
 class TypescriptSignaturePopup(sublime_plugin.TextCommand):
     def is_enabled(self):
-        return TOOLTIP_SUPPORT and is_typescript(self.view)
+        return TOOLTIP_SUPPORT and is_typescript(self.view) and get_language_service_enabled()
```
DanielRosenwasser: I think it might be worthwhile to create decorators for these; basically you'd be able to do something like

* `typescript/commands/signature.py`
    ```python
    from ..libs.guards import is_typescript, language_service_enabled;

    class TypescriptSignaturePopup(sublime_plugin.TextCommand):
        @g_is_typescript
        @language_service_enabled
        def is_enabled(self):
            return TOOLTIP_SUPPORT
    ```
* `typescript/libs/guards.py`

    ```python
    def language_service_enabled(func):
        def wrapped(*args):
            func(*args) and get_language_service_enabled()
        return wrapped

    def g_is_typescript(func):
        def wrapped(self, *args):
            func(self, *args) and is_typescript()
        return wrapped
    ```

zhengbli: That doesn't seem to have much gain over what it is right now though. It might be worthwhile to refactor all the labels later.

NekR:
Would be good to have this merged and released.

DanielRosenwasser:
:+1:

PR #411, opened 2015-12-02
Adding async/await syntax highlighting support.

ejsmith:


mhegazy:
@ejsmith can you send this PR to https://github.com/Microsoft/TypeScript-TmLanguage instead?

ejsmith:
Ugh, clicked the wrong repo.

PR #420, opened 2015-12-14
fix display of screenshots on packagecontrol.io

MattDMo:
The `packagecontrol.io` [`TypeScript` page](https://packagecontrol.io/packages/TypeScript) displays the screenshots immediately after the description in the bullet point, not on a new line:

with the first image displaying directly after the text instead of having each image at a uniform place and size. The second bullet text and image display fine because the text is longer, causing the image to wrap. Both lines display correctly on Github. This is because Github-flavored Markdown interprets a single newline in the `.md` source as a `<p>` in the HTML, whereas the original Markdown processor and derivatives require two newlines for a `<p>`. I haven't looked through wbond's source for page rendering, but it appears he doesn't have full GFM compliancy turned on.

At any rate, to make a short story long, I added newlines after each bullet point, so everything should line up nicely now both on Github and `packagecontrol.io`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,30 +83,39 @@ The plugin also supports representing a TypeScript project via a [tsconfig.json]
 Screenshots
 ------
 - Project error list
+
 ![](https://raw.githubusercontent.com/Microsoft/TypeScript-Sublime-Plugin/master/screenshots/errorlist.gif)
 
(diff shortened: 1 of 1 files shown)


DanielRosenwasser:
Thanks @MattDMo!

PR #421, opened 2015-12-16
Update README.md

aceslick911:
Two very important and useful features: auto-completion and signature pop-up should be described in the read-me.


DanielRosenwasser:
Thanks for the PR!

PR #432, opened 2016-01-28
Update TypeScript.tmLanguage

feliperohdee:
Include \s+ to class declaration, in case of object with 'class' property, it doesn't breaks color scheme


zhengbli:
Hi @feliperohdee , all changes to the tmLanguage files are located at https://github.com/Microsoft/TypeScript-TmLanguage, because they are shared among several different editors. So please redirect the contribution there, thank you!

feliperohdee:
Got it @zhengbli, thank u!

PR #433, opened 2016-02-01
Support customized typescript tsdk location

### LINKED ISSUE #255: Allow user to configure path to tsc and tsserver
For now we are shipping a TypeScript compiler and language service along with the plugin. However, some users have the tendency to stick with certain version of TypeScript and want to use that version in the plugin. It would be good to give them the ability to configure the TypeScript version they want to use.

**Current state**
Currently we have a setting "typescript_proc_file" to specify path to `tsserver.js`. There is no similar setting for tsc.

```
# retrieve the path to tsserver.js
# first see if user set the path to the file
settings = sublime.load_settings('Preferences.sublime-settings')
proc_file = settings.get('typescript_proc_file')
if not proc_file:
	# otherwise, get tsserver.js from package directory
	proc_file = os.path.join(PLUGIN_DIR, "tsserver", "tsserver.js")
```
			
**Propose**
There should be two settings "typescript_tsc_file" and "typescript_tsserver_file" with the following possible values:
- "plugin": use the tsc/tsserver shipped along with the plugin
- "global": use the tsc/tsserver in their PATH (most likely installed with NPM)
- path to executable tsc/tsserver (e.g. tsc.exe) or tsc.js/tsserver.js.

zhengbli:
Now it uses the same setting name `typescript_tsdk` with VSCode.

Fix #255

### DIFF
--- a/typescript/libs/editor_client.py
+++ b/typescript/libs/editor_client.py
@@ -49,8 +49,8 @@ class EditorClient:
 
         # retrieve the path to tsserver.js
         # first see if user set the path to the file
-        settings = sublime.load_settings('Preferences.sublime-settings')
-        tsdk_location = settings.get('typescript_tsdk')
+        settings = sublime.load_settings("Preferences.sublime-settings")
+        tsdk_location = settings.get("typescript_tsdk")
         if tsdk_location:
             proc_file = os.path.join(tsdk_location, "tsserver.js")
             global_vars._tsc_path = os.path.join(tsdk_location, "tsc.js")
--- a/typescript/libs/editor_client.py
+++ b/typescript/libs/editor_client.py
@@ -2,6 +2,7 @@
 from .node_client import ServerClient, WorkerClient
 from .service_proxy import ServiceProxy
 from .global_vars import *
+from . import global_vars
 
 
 class ClientFileInfo:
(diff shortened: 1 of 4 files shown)


typescript/libs/editor_client.py
```diff
@@ -49,11 +50,16 @@ def initialize(self):
         # retrieve the path to tsserver.js
         # first see if user set the path to the file
         settings = sublime.load_settings('Preferences.sublime-settings')
-        proc_file = settings.get('typescript_proc_file')
-        if not proc_file:
+        tsdk_location = settings.get('typescript_tsdk')
```
DanielRosenwasser: Quotes

mhegazy:
can you please update the steps in https://github.com/Microsoft/TypeScript/wiki/Nightly-drops#sublime-text after this is merged in.

typescript/libs/global_vars.py
```diff
@@ -34,7 +34,9 @@
 def get_node_path():
     return _node_path
 
-TSC_PATH = os.path.join(PLUGIN_DIR, "tsserver", "tsc.js")
+_tsc_path = None
```
DanielRosenwasser: There's a comment above that says where `_node_path` gets initialized. Can you add a similar one here?

zhengbli: Done

typescript/libs/editor_client.py
zhengbli: Done

zhengbli:
@mhegazy @DanielRosenwasser further comments?

DanielRosenwasser:
:+1:

PR #435, opened 2016-02-10
Add missing comma in list declaration, fix #413

### LINKED ISSUE #413: Tab completion causes source misinterpretation
This is on the latest version of the plugin with TypeScript 1.7.3 on OS X.

When working with completion on long variable names, a strange issue can pop up. If I work from a blank file and type out:

```js
let SomeLongName = "Hello, World";
console.log(SomeL <tab>
```
(where `<tab>` is pressing of the tab key to trigger autocomplete)

I'll end up with:
```js
let SomeLongName = "Hello, World!";
console.log(SomeLongName);
```

But `SomeLongName` will be squiggly-underlined, with the status bar showing an error:
```
Cannot find name 'SomeLongNameongName'., line 2, column 25
```

The error persists until I reopen the buffer or delete and retype the variable name without using completion.

All relevant editor settings in Sublime Text are set as their defaults.

cristiandima:
The missing comma makes the two distinct list items `"insert_best_completion"` and `"typescript_format_on_key"` concatenate into a single 
`"insert_best_completiontypescript_format_on_key"` list item. The if condition then evaluates to true even when command_name is one of the two names which is clearly not intended.

This change should fix #413. Tab completion emits the "insert_best_completion" command which eventually takes the interpreter to the aforementioned if block and has it call the `reload_buffer` function. I was no longer able to reproduce #413 after this change.

### DIFF
--- a/typescript/listeners/listeners.py
+++ b/typescript/listeners/listeners.py
@@ -220,7 +220,7 @@ class TypeScriptEventListener(sublime_plugin.EventListener):
         log.debug("on_post_text_command_with_info")
         if command_name not in \
             ["commit_completion",
-             "insert_best_completion"
+             "insert_best_completion",
              "typescript_format_on_key",
              "typescript_format_document",
              "typescript_format_selection",


billti:
Awesome find! Thanks for fixing. This has been a pain for a while.

zhengbli:
Oh man. Thanks for the finding @cristiandima !

PR #437, opened 2016-02-13
Used tab_size as indent_size value by default

MOZGIII:


zhengbli:
:+1: LGTM. Thanks for the contribution @MOZGIII !

PR #445, opened 2016-02-26
Autocomplete fix, this closes #369 and closes #442

### LINKED ISSUE #369: Snippets stop working in TypeScript files
Snippets seem to be clobbered by the most recent version of the plugin on PackageControl + NPM typescript install. It's possible I'm missing something - so please comment if it works for you. When I provide SublimeText (v3) a snippet (without a source scope) it works across other languages (JS, python, etc). Simple test case:

    <snippet>
        <content><![CDATA[
    class ${1:name} {
        constructor() {}
    }
    
    ]]></content>
        
        <tabTrigger>tsclass</tabTrigger>
        
        
    </snippet>

### LINKED ISSUE #442: Loss of Autocompletion?
Just installed this but seem auto completion is gone or lacking.
There's some autocomplete for keywords but none of the variables defined in .ts file 
Am I missing something?

Using: 
OSX 10.10
Using Sublime Build 3065
Installed this using the package control

silvermario:
This fixes the auto-complete issues in Sublime Text 3.

### DIFF
--- a/typescript/listeners/completion.py
+++ b/typescript/listeners/completion.py
@@ -144,7 +144,7 @@ class CompletionEventListener:
     def run_auto_complete(self):
         active_view().run_command("auto_complete", {
             'disable_auto_insert': True,
-            'api_completions_only': True,
+            'api_completions_only': False,
             'next_completion_if_showing': False,
             'auto_complete_commit_on_tab': True,
         })


zhengbli:
This will close #369, though #442 seems to be a different problem (the user cannot use intellisense in general).

zhengbli:
@silvermario Thanks for the contribution!

Issue #387, opened 2015-10-27
typescript_auto_format should be off by default


matthauck:
I was rather surprised to see sublime auto-formatting my code for me. This is a rather intrusive thing and I think this kind of setting should be off by default. I don't like editors that change my code without my telling it to, which is one of the reasons I like sublime, for its simplicity. 

If you decide to keep this on by default, please at least document this setting in your README so that users don't have to dig around to find the right sublime-settings file (since it doesn't show up under "Preferences -> Package Settings" for some reason) to figure out how to turn it off when sublime starts doing things they didn't expect it to do.

Thanks.

zhengbli:
I admit that autofomatting by default may cause confusion for some users, however the default formatting only changes the current line, which has limited impact. And the behavior is consistent with what Visual Studio has too.

I think it might be better to update the README to make users more informed about this behavior. I'll send a PR soon.

Issue #309, opened 2015-07-24
Copy+Paste broken for TS Files


danmane:
After installing this plugin, I find that paste is broken (only for Typescript files). Pressing cmd+v has no effect.  I see that the plugin has overwritten cmd+v to paste and format. It appears that this functionality is not working as intended.

danmane:
Actually, it may be that the typescript language services died entirely, which knocked out paste functionality as a side-effect.

hoanhtien:
Doesn't work indeed, even on Windows. This is a bug.

hoanhtien:
@danmane The reason it does not work on Windows is because we overwrite Cmd+v and Windows does not have a Cmd key. It is a separate problem to what you are having (I assume you are using MAC).
```
    {
        "keys": [ "super+v" ],
        "command": "typescript_paste_and_format",
        "context": [
            { "key": "setting.typescript_auto_format", "operator": "equal", "operand": true },
            { "key": "selector", "operator": "equal", "operand": "source.ts" }
        ]
    }
```
We check on a MAC and Cmd+V works without problem. Could you confirm that typescript language service server is dead (symptoms: error messages on the Python console, auto-completion does not work, format-document does not work)?

Also have you tried to fully quit Sublime (using the Quit command on the application menu) and restart? This will restart the typescript server.

A work-around is to disable auto format by setting `"typescript_auto_format": "false"` and use `Ctrl + T`, `Ctrl + F` to manually format the whole document.

danmane:
@hoanhtien I can confirm that when it stopped working for me, it was because the Typescript server was dead. I was using sublime tunneling over SSH to modify a typescript project on a remote server, and the connection was flaky.

Closing the issue.

mgiraldo:
@danmane how did you fix this? i have Cmd+V not working in my Mac also. i am compiling manually via `tsc -w`

mgiraldo:
never mind. i restarted Sublime and it worked again. thanks.

asimlqt:
This is definitely a bug. I have the same problem on Linux, Ctrl+V does not work in typescript files. Works fine in regular javascript files.

Disabling auto formatting as mentioned by @hoanhtien worked for me. Thanks.

shudson250:
Disabling auto formatting as mentioned by @hoanhtien worked for me as well.

To save you from digging through settings files, just add `"typescript_auto_format": false,` to
Package Settings > Typescript > Plugin Settings

[...]

Issue #195, opened 2015-05-12
typescript_auto_format breaks editorconfig's tab preference


jamiecounsell:
When `typescript_auto_format` is enabled, it appears that certain operations will cause indentation to be replaced with tabs, even when the .editorconfig file specifies the use of tabs. The easiest way to replicate it is to enter a new line that is indented by the auto-format, then paste a variable from the same file. This took me a while to figure out, I thought it was .editorconfig not working, but it only occurs with .ts files.

zhengbli:
This was an known issue before v1.4 release, but now it should work with sublime text's tab settings. 

And I don't quite understand the repro procedure you mentioned. Do you mean the following:
![195](https://cloud.githubusercontent.com/assets/1171301/7806035/bb538268-0330-11e5-9aa0-078088943d01.gif)
(Here the ``typescript_auto_format`` is enabled, and the ``indent using spaces`` sublime view setting  is set to true).

PR #454, opened 2016-03-05
Disable auto-format by default

zhengbli:
As required by many user-logged issues:
https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/387
https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/195
https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/309

It may be better to switch off the auto-formatting by default. This way even when the plugin failed to run, the basic copy / paste function wouldn't be broken.

### DIFF
--- a/Preferences.sublime-settings
+++ b/Preferences.sublime-settings
@@ -1,4 +1,4 @@
 {
-    "typescript_auto_format": true,
+    "typescript_auto_format": false,
     "enable_typescript_language_service": true
 }


billti:
What is the impact to the user experience? Is the default formatting with this off a good experience?

Also, why is the copy/paste broken if this is on and the plugin fails to run. Shouldn't the plugin be more resilient regardless?

zhengbli:
Personally I don't see much difference in the user experience, VSCode doesn't turn auto formatting on by default either. 

The reason why copy / paste is broken is that the "format on paste" command intercepts key-binding for the paste command, which does some pre-paste processing before explicitly calling the "paste" command inside the function. When the server is down, this paste command may not be called. We can do a try catch to make sure the paste command gets called regardless of the server status.

billti:
OK. Sounds good, and avoids some pain some users have hit. :+1: Thanks!

PR #467, opened 2016-04-17
Removes `$` from word_separators Fixes #457

### LINKED ISSUE #457: Completion list vanishes when typing $
Code:
```ts
var x = { $foo: 0 };
x. //<- type $ here
```
image

Result: When typing `$`, the completion list closes

Expected: Should stay up

subhaze:
Prevents '$' from breaking completions.
Also allows proper word selections when variables contain `$` within them.


zhengbli:
:+1: Thanks!

PR #513, opened 2016-09-08
Add primitive types null, true, and false

poke:
Starting with TypeScript 2, we can use all kinds of literal types. These three are very common literal types which also happen to be reserved words, so I guess it makes sense to handle them in a special way.

Ultimately, it might make sense to expand the syntax structure for literal types a bit.

### DIFF
--- a/TypeScript.tmLanguage
+++ b/TypeScript.tmLanguage
@@ -1850,7 +1850,7 @@
 				</dict>
 			</dict>
 			<key>match</key>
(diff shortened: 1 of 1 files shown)


mhegazy:
thanks @poke for the PR. The TM Language file is maintained in https://github.com/Microsoft/TypeScript-TmLanguage. this is shared between the sublime text plugin and other plugins like VSCode for instance. Please submit your PR to that repo instead.

poke:
Oh, alright, I’ll move over there then. Thanks for the quick response!

PR #527, opened 2016-09-24
Implement quick info tooltip on mouse hover

zhengbli:
With sublime build 3118, the new `on_hover` event enables us to show the quick info tooltip when hovering on a symbol.

Example:
sp160924_011345

### DIFF
--- a/typescript/commands/quick_info.py
+++ b/typescript/commands/quick_info.py
@@ -31,9 +31,7 @@ class TypescriptQuickInfoDoc(TypeScriptBaseTextCommand):
     re-runs quick info in case info has changed
     """
 
-    display_point = None
-
-    def handle_quick_info(self, quick_info_resp_dict):
+    def handle_quick_info(self, quick_info_resp_dict, display_point):
         if quick_info_resp_dict["success"]:
             info_str = quick_info_resp_dict["body"]["displayString"]
             status_info_str = info_str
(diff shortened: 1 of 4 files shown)


zhengbli:
@billti comments?

typescript/commands/quick_info.py
```diff
@@ -47,20 +50,24 @@ def handle_quick_info(self, quick_info_resp_dict):
...
+
+            # process display string
             if TOOLTIP_SUPPORT:
                 html_info_str = escape_html(info_str)
                 html_doc_str = escape_html(doc_str)
                 html = "<div>" + html_info_str + "</div>"
                 if len(doc_str) > 0:
                     html += "<div>" + html_doc_str + "</div>"
-                self.view.show_popup(html, location=-1, max_width=800)
+                self.view.show_popup(html, flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY, location=self.display_point, max_width=800)
```
billti: Does `HIDE_ON_MOUSE_MOVE_AWAY` have any impact if the popup was manually triggered? (i.e. not from hover)?

typescript/commands/quick_info.py
```diff
@@ -31,11 +31,14 @@ class TypescriptQuickInfoDoc(TypeScriptBaseTextCommand):
     re-runs quick info in case info has changed
     """
 
+    display_point = None
```
billti: Having this state out here, and setting it in the run method, then using it in a later method, is kinda ugly. Is there no way to pass it directly?

typescript/commands/quick_info.py
zhengbli: It gives the same behavior. So if the user triggered the tooltip using the keyboard shortcut, the tooltip will show up as normal; and then if he/she moves the cursor or press any key, the tooltip will go away. Considering quick info tooltip is normally not necessary to be consistently open, this behavior seems reasonable.

typescript/commands/quick_info.py
zhengbli: Indeed it does smell. Removed.

PR #535, opened 2016-09-30
Fix #461: Add fallback to default paste command

### LINKED ISSUE #461: Cannot paste in unsaved TypeScript file
The TypeScript plugin overwrites the <kbd>Ctrl+V</kbd> keyboard shortcut to call the `typescript_paste_and_format` command instead. However, as per the logic in [base_command.py](https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/master/typescript/commands/base_command.py), these commands can be disabled.

This is usually not a problem when the TypeScript commands all require e.g. a path to resolve dependencies and things, however if the `typescript_paste_and_format` command fails to run, then the user cannot paste content, leading to a very frustrating experience.

So for new and unsaved files, the commands *are* disabled. This is because the [`view_helpers.is_typescript`](https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/master/typescript/libs/view_helpers.py#L99) check goes against the view’s file name. If there is no filename (which is the case when the file is new), then the function returns false, and the command gets disabled.

So this is not really an ideal check. Ideally, the commands should be enabled and do as much as it can that does not require an actual file path. For formatting, one would think that this should work even without a physical file. And even if it was not possible, the command should delegate back to the default paste command in that case.

So there are multiple ways to fix this (which is the reason why this is not a pull request):

- Change the `is_typescript` to return true when the file is a typescript file that was not saved. This could potentially break commands that really depend on a file path.
- Split up the `is_typescript` into `is_typescript` and `has_filepath` or something and change the commands so they only require what they actually require (a paste and format could then only require typescript syntax, while auto completion from imports may require paths). This would be the ideal solution as it makes the command usable as soon as they can, but it could be a bit tedious to implement, especially when there’s an overlap in command’s functionality.
- Just fix the paste and format command, so it overrides the `is_enabled` and properly runs even when the file is unsaved.
- Modify the paste and format command so it can always run but delegates back to the default paste command when the `is_typescript` returns false.

poke:
I just found an open tab in my browser about issue #461 and realized that I never got around to do this… So, here we go! This actually turned out to be simpler than what I originally expected to do.

This change basically always enables the `typescript_paste_and_format` command but moves the check whether the TypeScript formatter could run inside the execution. If it is not possible, then it falls back to the default `paste` command.

That way, we can safely overwrite the standard keyboard shortcut without breaking paste when TypeScript cannot do this. So we can happily paste into new buffers again—yay!

### DIFF
--- a/typescript/commands/format.py
+++ b/typescript/commands/format.py
@@ -71,7 +71,17 @@ class TypescriptFormatBrackets(TypeScriptBaseTextCommand):
 
 
 class TypescriptPasteAndFormat(TypeScriptBaseTextCommand):
+    def is_enabled(self):
+        return True
+
     def run(self, text):
+        if is_typescript(self.view) and get_language_service_enabled():
+            self._run(text)
+        else:
+            # fall back to default paste command
+            self.view.run_command('paste')
+
+    def _run(self, text):
         log.debug("running TypescriptPasteAndFormat")
         view = self.view
         check_update_view(view)


zhengbli:
👍  Thanks for the contribution!

poke:
That was super fast, thanks! :D

PR #546, opened 2016-10-12
Support inline error tooltip at cursor position

zhengbli:
The added behavior is quite similar to what we had before. But when the sublime version supports Phantom APIs (> 3118), we show a phantom popup instead of put the error message at the status bar (**only after you click**). The popup will go away when either the error is fixed or when the cursor (**the keyboard cursor** not the mouse cursor) moved to somewhere else. While the "on_hover" tooltip will always show the quick info.

In summary, the workflow is:

* hover with your mouse cursor to get quick info
* click at an error position to see error messages
* errors messages will appear when keyboard cursor is on top of an error span
* click elsewhere to dismiss an error message

gif

### DIFF
--- a/typescript/commands/error_info.py
+++ b/typescript/commands/error_info.py
@@ -24,8 +24,7 @@ class TypescriptErrorInfo(TypeScriptBaseTextCommand):
 
         if len(error_text) > 0:
             if PHANTOM_SUPPORT:
-                template = '<body><style>div.error {{ background-color: brown; padding: 5px; }}</style><div class="error">{0}</div></body>'
+                template = '<body><style>div.error {{ background-color: brown; padding: 5px; color: white }}</style><div class="error">{0}</div></body>'
                 display_text = template.format(error_text)
                 self.view.add_phantom("typescript_error", self.view.sel()[0], display_text, sublime.LAYOUT_BLOCK)
-            else:
-                self.view.set_status("typescript_error", error_text)
+            self.view.set_status("typescript_error", error_text)
--- a/typescript/commands/error_info.py
+++ b/typescript/commands/error_info.py
@@ -21,7 +21,11 @@ class TypescriptErrorInfo(TypeScriptBaseTextCommand):
             if region.contains(pt):
                 error_text = text
                 break
+
         if len(error_text) > 0:
-            self.view.set_status("typescript_error", error_text)
-        else:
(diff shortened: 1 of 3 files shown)


typescript/commands/error_info.py
```diff
@@ -21,7 +21,11 @@ def run(self, text):
...
                 error_text = text
                 break
+
         if len(error_text) > 0:
-            self.view.set_status("typescript_error", error_text)
-        else:
-            self.view.erase_status("typescript_error")
\ No newline at end of file
+            if PHANTOM_SUPPORT:
+                template = '<body><style>div.error {{ background-color: brown; padding: 5px; }}</style><div class="error">{0}</div></body>'
```
billti: Does the coloring change at all with the themes? Seeing as you only set background here, and not foreground, could foreground ever default to something in some themes which is unreadable?

typescript/commands/error_info.py
```diff
@@ -21,7 +21,11 @@ def run(self, text):
...
+                display_text = template.format(error_text)
+                self.view.add_phantom("typescript_error", self.view.sel()[0], display_text, sublime.LAYOUT_BLOCK)
```
billti: Interesting I don't see `add_phantom` on the `view` class in the docs (https://www.sublimetext.com/docs/3/api_reference.html#sublime.View). Maybe an oversight. They didn't make changes to this API after introducing it right?

typescript/commands/error_info.py
```diff
@@ -21,7 +21,11 @@ def run(self, text):
             if region.contains(pt):
                 error_text = text
                 break
+
         if len(error_text) > 0:
-            self.view.set_status("typescript_error", error_text)
-        else:
-            self.view.erase_status("typescript_error")
\ No newline at end of file
+            if PHANTOM_SUPPORT:
```
billti: If `show_errors_inline` is set to `false`, do phantoms still appear? As this is an `if/else`, that would mean no error info. Is there a downside to always updating the status bar anyway?

zhengbli: Yes, `show_errors_inline` only affects sublime's own error tips, it doesn't toggle all phantom tooltips. Though I agree we can always update the status bar anyway.

billti:
👍

camloken:
It would be cool if you could place links in the tooltip that take you to the website rule. Atom does this.

PR #559, opened 2016-11-02
make error tips optional

zhengbli:
Per popular request this design may get in the way of people's working flow. Making it optional.

### DIFF
--- a/typescript/listeners/idle.py
+++ b/typescript/listeners/idle.py
@@ -169,7 +169,7 @@ class IdleListener:
     def update_status(self, view, info):
         """Update the status line with error info and quick info if no error info"""
         # Error info
-        if is_inline_error_tips_enabled():
+        if PHANTOM_SUPPORT:
             view.erase_phantoms("typescript_error")
         view.erase_status("typescript_error")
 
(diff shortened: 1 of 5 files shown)


typescript/listeners/idle.py
```diff
@@ -169,7 +169,7 @@ def show_errors(self, diagno_event_body, syntactic):
     def update_status(self, view, info):
         """Update the status line with error info and quick info if no error info"""
         # Error info
-        if PHANTOM_SUPPORT:
+        if is_inline_error_tips_enabled():
```
DanielRosenwasser: Is this the right check? I feel like you could get into a bad state if you

1. Enable inline error tips
2. Cause an error.
3. Disable inline error tips.

After 3, will this erase ever occur? I'm not familiar with the API, but I'm not sure it will. I think it's probably better to always erase based on `PHANTOM_SUPPORT`.

zhengbli: Good point, I've updated.

gcnew:
May you publish a new version to Package Control as well? I cloned the repo, but maybe not all users will do it.

Mesh066Z:
@gcnew

Mesh066Z:
That's nice for the workers to assess their job immediately after they have submitted the paperwork

PR #562, opened 2016-11-18
Combine error tooltip with quick info

zhengbli:
We learned from user feedback that the initial implementation of the inline error tooltip is annoying and gets in the way of typing. This PR changes the implementation to show the error tip during mouse hovering, which is the same behavior with VSCode and VS.

Now the quick info hover tooltip is combined with the error tooltip, so if there is no error in the position, only quick info should be shown:
image
otherwise both will be shown:
image


roblav96:
@zhengbli I like this one :D

EricNilsson:
+1 :D

typescript/commands/quick_info.py
```diff
@@ -49,17 +56,31 @@ def handle_quick_info(self, quick_info_resp_dict, display_point):
...
+            self.view.show_popup(html, flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY, location=display_point, max_height=300, max_width=1500)
+
+    def get_error_text_html(self, pt):
+        client_info = cli.get_or_add_file(self.view.file_name())
+        error_text = ""
+        for (region, text) in client_info.errors['syntacticDiag']:
+            if region.contains(pt):
+                error_text = text
+                break
+        for (region, text) in client_info.errors['semanticDiag']:
```
DanielRosenwasser: One problem I notice with this approach is that it only handles one error at a time. Also, it requires a full iteration of all the diagnostics. But we could fix this another time.

DanielRosenwasser:
Looks good apart from what I can see (apart from nits).

Issue #518, opened 2016-09-13
Word multiple selection with pressed ctrl don't work 


BjornMelgaard:
In all document types on pressed ctrl + double click first word select normally, on next dbclick sublime remove all selections, but must select whole word. 
After removing through Package control all works fine.
Linux, Sublime 3 build 3114

BjornMelgaard:
https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/514

With old mousemap dbclick without ctrl don't select word as expected, with new - dbclick+ctrl don't select second and subsequent words as expected, delete this mousemap at all

richterdennis:
This bug is so annoying, please fix this as soon as possible

alexburner:
Confirmed, issue #514 was happening to me and was very confusing.
Also confirmed, pull #515 fixed 514 but added the new bug described here.

richterdennis:
Inspired by #514 and #515, my tmp fix is, adding the `alt` key to the modifiers

`Default.sublime-mousemap`:
```
[
    {
        "button": "button1",
        "count": 2,
        "modifiers": ["ctrl", "alt"],
        "command": "typescript_go_to_ref",
        "press_command": "drag_select"
    }
]
```

alexburner:
Yesss @richterdennis that worked for me, thanks!

For anyone else hunting, the file is in `path/to/Packages/TypeScript/Default.sublime-mousemap`

SystemDisc:
This issue is very irritating. It took me a bit to figure out which package was responsible.

donmckenna:
@richterdennis Thank you, works for me as well!

Narretz:
@zhengbli Can we get this issue triaged? This is really annoying since it interferes with core Sublime functionality.

stylemistake:
Recently found that this package was the source of this behaviour. Extremely annoying.

alexburner:
Bump. Just had to re-hunt this down and patch it again. For my future self, run this:
```
$ subl ~/.config/sublime-text-3/Packages/TypeScript/Default.sublime-mousemap
```

alexburner:
Adding a PR to make the patch from @richterdennis official

alexburner:
PR #567 has been merged and this issue should be fixed, but I don't have any sort of permissions to close.

Issue #514, opened 2016-09-08
Cannot double click to select word in any language after recent update


gurutech-thomasr:
Recent commit globally disables basic text editing feature in Sublime Text 3 across all languages.  New `Default.sublime-mousemap` is culprit, binding’s `context` is suspect:

```
Revision: 4582a1542dda4848805398713cde9505be5d01c3
Author: zhengbli <zhengbli@microsoft.com>
Date: 2016-09-02 17:13:07
Message:
Add mouse click support for go to ref


----
Added: Default.sublime-mousemap
```
```json
[
    {
        "button": "button1",
        "count": 2,
        "command": "typescript_go_to_ref",
        "press_command": "drag_select",
        "context": [
            { "key": "selector", "operator": "equal", "operand": "text.find-refs" }
        ]
    }
]
```

Steps to reproduce:

1. Open `Packages\TypeScript\Default.sublime-mousemap`.
2. Open any other text file.
3. Double click on a word does not select the word as expected.
4. Truncate `Default.sublime-mousemap` and save.
5. Double click on word now selects the word as expected.

Workaround:  Delete `Default.sublime-mousemap` or provide user-specific alternative.

charleylla:
it works! thank you very much!

PR #567, opened 2016-12-01
Fix word multi select while holding "ctrl"

alexburner:
In Sublime Text (PC/Linux) holding the `ctrl` key allows you to create multiple simultaneous selections with your mouse.

As tracked in issue #518, the fix here from #514 breaks this core functionality.

Adding the `alt` key to this modifiers array (as suggested by @richterdennis) prevents the conflict.

### DIFF
--- a/Default.sublime-mousemap
+++ b/Default.sublime-mousemap
@@ -2,7 +2,7 @@
     {
         "button": "button1",
         "count": 2,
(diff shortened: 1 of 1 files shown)


DanielRosenwasser:
Thanks @alexburner!

alexburner:
Hooray! Glad I could help a little

PR #570, opened 2016-12-09
Add language support for JavaScript

zhengbli:
Until now this plugin only handles .ts files and .tsx files. This PR adds supports for .js files and .jsx files as well.


typescript/libs/view_helpers.py
```diff
@@ -106,8 +106,9 @@ def is_typescript(view):
     except:
         return False
 
-    return (view.match_selector(location, 'source.ts') or
-            view.match_selector(location, 'source.tsx'))
+    is_ts_file = view.match_selector(location, 'source.ts') or view.match_selector(location, 'source.tsx')
+    is_js_file = view.match_selector(location, 'source.jsx') and cli.enable_language_service_for_js
```
DanielRosenwasser: Does Sublime consider the entire region of a .js file to be a `source.jsx` scope? Or is this from supplying our own `tmLanguage`? Can you add a comment explaining, since it's not obvious why you're not checking for `scope.js`.

TypeScript.sublime-settings
```diff
@@ -1,5 +1,5 @@
 {
-    "auto_complete_triggers" : [ {"selector": "source.ts", "characters": "."} ],
+    "auto_complete_triggers" : [ {"selector": "source.ts", "characters": "."}, {"selector": "source.jsx", "characters": "."} ],
```
hoest: Better change this to `source.js` and add this selector to the `TypeScriptReact.sublime-settings` file.

typescript/libs/view_helpers.py
hoest: I think adding the selector for `source.js` beside `source.jsx` would be better. Like mentioned in my other comment.

batusai513:
Any updates on this one?

typescript/libs/view_helpers.py
idiotWu: I am using `view.match_selector(location, 'source.ts, source.tsx, source.js, source.jsx')` here and it works perfectly.

idiotWu:
I think some key trigger bindings should also be adjusted for JavaScript files, eg:

```diff
{
  "keys": ["("],
  "command": "typescript_signature_popup",
  "context": [
-      { "key": "selector", "operator": "equal", "operand": "source.ts, source.tsx" },
+      { "key": "selector", "operator": "equal", "operand": "source.ts, source.tsx, source.js, source.jsx" },
      { "key": "paren_pressed" }
  ]
}
```

Besides, I'm just wondering if this PR will be merged soon 😅 . Type completion for JavaScript files works perfectly in vscode, so I (and many other users) are looking forward to use this feature in sublime.

james2doyle:
Since the issue seems to have gone stale, I had to turn to the Sublime LSP plugin and a different language server to get plain `.js` files working properly.

[You can read more about how I got this working on the LSP Plugin repo](https://github.com/tomv564/LSP/issues/369).

ryanolsonx:
Ping

james2doyle:
I think this is closed now due to #620

PR #610, opened 2017-06-24
Avoid using the ambiguous \h shorthand character

pchaigno:
Depending on the regular expression engine used, \h does not always
mean the same. With a PCRE engine, it matches white spaces, whereas,
with a Oniguruma engine, it matches hexademical digit characters.
Sublime Text uses an Oniguruma engine, but github.com relies on a
PCRE engine.

### DIFF
--- a/TypeScript.tmLanguage
+++ b/TypeScript.tmLanguage
@@ -4901,7 +4901,7 @@
                 <key>name</key>
                 <string>constant.other.character-class.range.regexp</string>
                 <key>match</key>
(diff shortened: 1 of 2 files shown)


pchaigno:
@zhengbli @DanielRosenwasser Did you get a chance to look into this?

zhengbli:
@pchaigno Hi thanks for contributing! Sorry I left Microsoft recently, so am not actively contributing to this project. But all tmlanguage changes should go to https://github.com/Microsoft/TypeScript-TmLanguage, this repo merely takes the latest from there. In addition, the .tmlanguage files are generated from the .YAML-tmLanguage files with the same name, they are not supposed to be modified by hand. 

The changes themselves look good to me though.

pchaigno:
Thanks for your answer.

Does that mean that `.tmLanguage` files are deprecated in this repo, or are they synchronized with the other repository? If they're synchronized, how often?

zhengbli:
The tmLanguage files are a bit outdates, and we used to adopt the latest from the tmLanguage repo when we draft releases, so not quite often.

Issue #561, opened 2016-11-12
Disable Linting


wildeyes:
Hello, is it possible to stop linting so I could use sublimelinter instead?

zhengbli:
If you mean disabling language service, you can do set 'enable_typescript_language_service' to false in your sublime settings.

wildeyes:
That disables all language services... Not just linting ...

zhengbli:
What features do you mean by linting?

wildeyes:
Linting is showing errors about the code in the editor. We can divide
showing errors in typescript into at least two categories, errors about bad
code, like unused variables, and errors about incorrect code, like type
errors caught by the compiler.

I would like an option (ideally) to disable each, but would settle for a
flag to disable both.

zhengbli:
You can specify how strict the error checking is by setting things in your tsconfig.json. https://www.typescriptlang.org/docs/handbook/compiler-options.html

For example, you can set `noUnusedLocals` to false so that will not be warned. But there are no ways currently to just disable error checking while still have the rest of the language service running.

wildeyes:
I don't want to disable globally typechecking, I want to use
sublime-linters functionality.

I know it doesn't currently exist. That's why I opened the issue. This
issue is a feature request.

On 12 Apr 2017 00:31, "Zhengbo Li" <notifications@github.com> wrote:

You can specify how strict the error checking is by setting things in your
tsconfig.json. https://www.typescriptlang.org/docs/handbook/compiler-
options.html

For example, you can set noUnusedLocals to false so that will not be
warned. But there are no ways currently to just disable error checking
while still have the rest of the language service running.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/Microsoft/TypeScript-Sublime-Plugin/issues/561#issuecomment-293406182>,
or mute the thread

zhengbli:
Fair enough, I'm reopening the issue.

Though I still don't quite understand what you mean by "using sublime linter". What is stopping you from using it today? Are there conflicts between these two plugins?

wildeyes:
I'm seeing the same errors twice.

Issue #550, opened 2016-10-25
Errors highlighting color


rbaumi:
Is it possible to change the color (or maybe the style) of errors highlighting? What I mean is the color of the `Keywords` from the theme.

![](https://s16.postimg.org/9jywfs5ud/Screenshot_from_2016_10_25_22_46_01.png)

Any way to change the color and / or mark it on the line number (like linter does)?

rbaumi:
Ok, I could manage to do it by changing lines 164 in file [typescript/listeners/idle.py](https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/master/typescript/listeners/idle.py) to:

```py
view.add_regions(region_key, error_regions, "invalid.illegal", "",
    sublime.DRAW_SOLID_UNDERLINE)
```

It would be good if both color and mark style was an config option. Also would be nice to be able to allow for marking the whole line (`start_offset` variable).

zhengbli:
Thanks for nivestigating! You are always welcome to send a PR :)

alexburner:
Thanks for the patch @rbaumi, I can see errors now!

rgant:
A work around for this is to add this to your .tmTheme in Sublime:

```xml
<dict>
  <key>name</key>
  <string>TypeScript Errors</string>
  <key>scope</key>
  <string>keyword</string>
  <key>settings</key>
  <dict>
    <key>foreground</key>
    <string>#F92672</string>
  </dict>
</dict>
<dict>
  <key>name</key>
  <string>Keyword</string>
  <key>scope</key>
  <string>keyword.control, keyword.operator, keyword.other</string>
  <key>settings</key>
  <dict>
    <key>foreground</key>
    <string>#cc99cc</string>
  </dict>
</dict>
```

betorobson:
@rgant these change on .tmTheme should display erros with gutter and outline?
I have tried it but I got any change :-(

rgant:
I don't use GitGutter (weird I know) so I'm not 100% what you are talking about. My styles above only apply to the Sublime view of the file.

<img width="362" alt="screen shot 2018-11-08 at 10 14 17" src="https://user-images.githubusercontent.com/710553/48207685-28196d80-e33f-11e8-97f9-70b4a1fd7ae7.png">

The pink squiggles in the screen shot above are scope "keyword" and are from TypeScript's error checking. The purple "import" and "from" are scope "keyword.control" (or "keyword.operator" or "keyword.other").

[...]

Issue #522, opened 2016-09-20
Show errors in gutter?


richardm:
Is it possible to show TypeScript errors in the Sublime gutter? Other plugins such as linters do this, and it makes it much easier to identify and fix errors.

abenhamdine:
why did you closed this one ? It's indeed an excellent idea.

Issue #478, opened 2016-05-21
Setting for error display mode


Xavdidtheshadow:
Hey there!

I was wondering if there was a setting for how linting errors are displayed inline (or if there isn't, if we could add one). 

a ts error: 
<img width="515" alt="screen shot 2016-05-21 at 1 34 41 am" src="https://cloud.githubusercontent.com/assets/1231935/15447352/88c305f4-1ef4-11e6-97fb-2aa38c4b97a8.png">

This is fairly quiet (you have to be looking at that line to see the little red squiggle). Contrast this with standard's [sublime-linter](https://github.com/Flet/SublimeLinter-contrib-standard): 

<img width="431" alt="screen shot 2016-05-21 at 1 35 39 am" src="https://cloud.githubusercontent.com/assets/1231935/15447364/bde151fa-1ef4-11e6-9e9b-d427451ee923.png">

The dot + box is much more visible. 

I'm not sure the scope of the work (if it's as simple as adding a setting to change how invalid lines are marked vs a large code change), but I think it'd be worth looking into!

zhengbli:
PRs are welcomed. These lines can be a good start:

https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/2da61f59d231b5442e1a7136aedcb9705ad099e2/typescript/listeners/idle.py#L164

PR #625, opened 2017-10-08
Increase error visibility

PSeitz:
Add gutter, outline errors

### DIFF
--- a/typescript/listeners/idle.py
+++ b/typescript/listeners/idle.py
@@ -158,13 +158,13 @@ class IdleListener:
 
         # Highlight error regions in view
         if IS_ST2:
-            view.add_regions(region_key, error_regions, "keyword", "",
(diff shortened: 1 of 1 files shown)


PSeitz:
#522 #550 #478

Generally it would be the right thing to extend SublimeLinter 3 (http://www.sublimelinter.com/en/latest/index.html) instead of recreating the SublimeLinter behaviour to get all the nice defaults user expect from Sublime Linting (like disabling #561)

betorobson:
do you have screenshot how it looks like?
You have my vote to aprove this PR!!!

betorobson:
Any update to get erros gutters and output log?
The tooltip mouse over is not helping at all.

orta:
This is deprecated by https://github.com/microsoft/TypeScript-Sublime-Plugin/pull/715

Issue #618, opened 2017-09-11
Stop spamming ERROR: can not send request; node process not running, on my console


evandrocoan:
I have this package installed on my Sublime Text on a machine where I do not have `node` installed. So it is expected it to not work, however this should not keep spamming my console with these messages every time some thing happens:
```
unloading plugin MaxPane.layouts_doc
unloading plugin MaxPane.max_pane
unloading plugin MultiEditUtils.MultiEditUtils
unloading plugin MultiEditUtils.selection_fields
unloading plugin UnitTesting.ut
2017-09-11 11:41:29,232: -1219987712: ERROR: can not send request; node process not running
2017-09-11 11:41:29,241: -1219987712: ERROR: can not send request; node process not running
2017-09-11 11:41:29,251: -1219987712: ERROR: can not send request; node process not running
2017-09-11 11:41:29,261: -1219987712: ERROR: can not send request; node process not running
2017-09-11 11:41:29,267: -1219987712: ERROR: can not send request; node process not running
2017-09-11 11:41:29,282: -1219987712: ERROR: can not send request; node process not running
reloading settings Packages/User/Package Control.sublime-settings
Unable to parse binding {keys: [ctrl+alt+w]}
Unable to parse binding {keys: [ctrl+shift+w]}
Unable to parse binding {keys: [ctrl+alt+w]}
Unable to parse binding {keys: [ctrl+shift+w]}
reloading settings Packages/User/Preferences.sublime-settings
2017-09-11 11:41:31,854: -1219987712: ERROR: can not send request; node process not running
2017-09-11 11:41:31,865: -1219987712: ERROR: can not send request; node process not running
2017-09-11 11:41:31,877: -1219987712: ERROR: can not send request; node process not running
2017-09-11 11:41:31,886: -1219987712: ERROR: can not send request; node process not running
2017-09-11 11:41:31,889: -1219987712: ERROR: can not send request; node process not running
2017-09-11 11:41:31,892: -1219987712: ERROR: can not send request; node process not running
Package Control: The dependency 'python-markdown' is installed, but out of date; upgrading to latest release 2.6.9 from 2.6.6...
Package Control: The dependency 'python-markdown' has successfully been installed or updated
Package Control: Upgraded GitGutter to 1.7.4
Package Control: Upgraded MaxPane to 2017.09.06.17.38.05
[...]

PR #648, opened 2018-03-04
Stop spamming ERROR: can not send request; node process not

evandrocoan:
running, on console

Fix Microsoft/TypeScript-Sublime-Plugin#618 Stop spamming ERROR: can not send request; node process not running, on my console


orta:
I think keeping this as an error makes sense (I think most people who would see this probably want to know that their node isn't configured correct) but I'm open to the PR making sure it only posts it once instead?

evandrocoan:
If their installation is not working they should enable the debug mode.

Issue #604, opened 2017-05-03
GoTo Definition shows all function calling in file


vosecek:
GoTo Definition by cmd+r on Mac shows all function calling in open file of project. The tool is useless now, if there are multiple calling of one method in current file.

![screenshot 2017-05-03 11 18 14](https://cloud.githubusercontent.com/assets/3978659/25657908/70e9661c-3000-11e7-97cf-146ac1428e2e.png)

billti:
Is this only for the Sublime "GoTo definition"? Does the TypeScript one (F12) work as expected for you?

longitude:
This actually happens in 'Go To Symbol' which navigates across symbols in a given file, (F12) 'Go To Definition' works as expected

Saskoh:
Here's a workaround to display only method names in the symbol list by setting up a configuration file for SublimeText that excludes bindables, method calls, etc.:

1) Open the SublimeText Packages Folder and navigate to the "TypeScript" folder
2) Inside the "TypeScript" folder create a new file called "Symbol List.tmPreferences"
3) Edit this file and copy the following lines into it:

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
  <key>name</key>
  <string>Symbol List</string>
  <key>scope</key>
  <string>meta.decorator.ts, meta.var.expr.ts, meta.block.ts, meta.field.declaration.ts</string>
  <key>settings</key>
  <dict>
    <key>showInSymbolList</key>
    <integer>0</integer>
  </dict>
</dict>
</plist>
```

4) Save the file

The next time you press "ctrl-r" in a TypeScript file, you'll only get the method names displayed.

vosecek:
@Saskoh Thanks, works great.

stweedie:
Why has this not made it to the actual package?

BinaryNate:
@stweedie , I think it's just that nobody had created a PR yet. So, I just put up #653 for it.

kylebebak:
For anyone that stumbles on this, you can create a `.tmPreferences` file in any directory from which `.tmPreferences` files are loaded, included the `User` directory.

It can also have any name you want.

[...]

PR #653, opened 2018-03-29
Fix issue #604 - GoTo Definition shows all function calling in file

BinaryNate:
This is a fix for issue #604 .

I added a 'Symbol List.tmPreferences' file to fix the issue where if you try to go to a function definition, it takes you to the first instance of where the function is used (not necessarily its definition).

I have been using this fix locally for a few weeks now, but this file was removed when I updated my TypeScript plugin.

### DIFF
--- a/Symbol List.tmPreferences
+++ b/Symbol List.tmPreferences
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+  <key>name</key>
+  <string>Symbol List</string>
+  <key>scope</key>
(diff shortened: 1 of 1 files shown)


DanielRosenwasser:
Can you explain where this is documented, and which declarations this is intended to cover? For example, I don't *think* decorators should be considered definition sites.

stweedie:
Symbols [(From sublime's documentation)](http://docs.sublimetext.info/en/latest/reference/symbols.html).

The expressions list is pulled from [TypeScript.tmLanguage](https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/master/TypeScript.tmLanguage). the 'showInSymbolList' set to 0 means that the above expressions DO NOT appear in the symbols list. In this case, syntax equates to: don't show decorators, local variable declarations, blocks, or member variables.

The main goal is to remove some things from the symbols list. Right now, everything is shown, including local variable expressions and ALL function calls (not just the function definitions). Whichever meta elements in the .tmLanguage file you would like to ignore, place them in that list.

BinaryNate:
Closing this, since I switched to VS Code

PR #658, opened 2018-04-14
add organize imports command

it6:


typescript/libs/service_proxy.py
```diff
@@ -123,6 +123,22 @@ def format_on_key(self, path, location=Location(1, 1), key=""):
...
+                "type": "file",
+                "args": {
+                    "file": path
+                }
+            },
+        }
+        req_dict = self.create_req_dict("organizeImports", args)
+        json_str = json_helpers.encode(req_dict)
+        response_dict = self.__comm.sendCmdSync(json_str, req_dict["seq"])
+        if self.__worker_comm.started():
```
DanielRosenwasser: Do you have any idea why we have to do this in some requests but not others? @billti

DanielRosenwasser:
Nice! ⭐️ I'll try to review this soon, probably during the week.

typescript/libs/service_proxy.py
it6: I am new to this :), but I think this is what is happening

Commands that change the file (edit/close/open etc..) would need this extra worker thread for faster execution. When I tried it without extra worker I noticed significant delay.

Looking from other requests, this is used along with most file manipulation commands.

billti: I'm not sure. Seems odd to send it twice (line 137 & 139). Unless maybe it's if we have the extra process running for error checking/project wide errors, and it's to keep that process in sync with any changes (which would make sense with @it6 's comment "Commands that change the file..."). @zhengbli wrote that stuff.

PR #659, opened 2018-04-16
remove file path from rename popup

it6:
Before 

screen shot 2018-04-15 at 8 53 40 pm

After

screen shot 2018-04-15 at 9 12 18 pm


DanielRosenwasser:
Ideally we'd have just used `displayName`, but seems like a reasonable change either way. Thank you!

PR #661, opened 2018-04-17
add code fixes command

### LINKED ISSUE #630: Request to add auto import
Can you please add ability to auto import.
Ideally a `keyboard shortcut/menu item` which works on the `class` where the cursor is located.

it6:
fixes #630

### DIFF
--- a/TypeScript.sublime-commands
+++ b/TypeScript.sublime-commands
@@ -7,7 +7,7 @@
  { "caption" : "TypeScript: Format Selection", "command": "typescript_format_selection" },
  { "caption" : "TypeScript: Format Document", "command": "typescript_format_document" },
  { "caption" : "TypeScript: Find References", "command": "typescript_find_references" },
- { "caption" : "TypeScript: Request Code Fixes", "command": "typescript_get_code_fixes" },
+ { "caption" : "TypeScript: Request Code Fixes", "command": "typescript_request_code_fixes" },
  { "caption" : "TypeScript: Navigate To Symbol", "command": "typescript_nav_to" },
  { "caption" : "TypeScript: Rename", "command": "typescript_rename" },
  { "caption" : "TypeScript: Paste And Format", "command": "typescript_paste_and_format" },
--- a/TypeScript.sublime-commands
+++ b/TypeScript.sublime-commands
@@ -7,7 +7,7 @@
  { "caption" : "TypeScript: Format Selection", "command": "typescript_format_selection" },
  { "caption" : "TypeScript: Format Document", "command": "typescript_format_document" },
  { "caption" : "TypeScript: Find References", "command": "typescript_find_references" },
- { "caption" : "TypeScript: Code Fixes", "command": "typescript_get_code_fixes" },
+ { "caption" : "TypeScript: Request Code Fixes", "command": "typescript_get_code_fixes" },
  { "caption" : "TypeScript: Navigate To Symbol", "command": "typescript_nav_to" },
  { "caption" : "TypeScript: Rename", "command": "typescript_rename" },
  { "caption" : "TypeScript: Paste And Format", "command": "typescript_paste_and_format" },
--- a/TypeScript.sublime-commands
+++ b/TypeScript.sublime-commands
@@ -7,6 +7,7 @@
  { "caption" : "TypeScript: Format Selection", "command": "typescript_format_selection" },
  { "caption" : "TypeScript: Format Document", "command": "typescript_format_document" },
  { "caption" : "TypeScript: Find References", "command": "typescript_find_references" },
+ { "caption" : "TypeScript: Code Fixes", "command": "typescript_get_code_fixes" },
  { "caption" : "TypeScript: Navigate To Symbol", "command": "typescript_nav_to" },
  { "caption" : "TypeScript: Rename", "command": "typescript_rename" },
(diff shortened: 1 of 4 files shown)


TypeScript.sublime-commands
```diff
@@ -7,6 +7,7 @@
  { "caption" : "TypeScript: Format Selection", "command": "typescript_format_selection" },
  { "caption" : "TypeScript: Format Document", "command": "typescript_format_document" },
  { "caption" : "TypeScript: Find References", "command": "typescript_find_references" },
+ { "caption" : "TypeScript: Code Fixes", "command": "typescript_get_code_fixes" },
```
DanielRosenwasser: `TypeScript: Request Code Fixes`

typescript/libs/service_proxy.py
```diff
@@ -272,3 +272,40 @@ def create_req_dict(self, command_name, args=None):
...
+    def get_errors_in_file(self, path):
+        args = {
+            "file": path
+        }
+        req_dict = self.create_req_dict("semanticDiagnosticsSync", args)
+        json_str = json_helpers.encode(req_dict)
+        response_dict = self.__comm.sendCmdSync(json_str, req_dict["seq"])
+        return response_dict
+
+
```
DanielRosenwasser: Remove extra newline

typescript/libs/service_proxy.py
```diff
@@ -272,3 +272,40 @@ def create_req_dict(self, command_name, args=None):
...
+      column = cursor[1] + 1
+      if errors_in_file['success']:
+          errors_at_cursor = list(filter(lambda error:
+              error['start']['line'] == line  and
+              error['end']['line'] == line and
+              error['start']['offset'] <= column and
+              error['end']['offset'] >= column , errors_in_file['body']))
+          return errors_at_cursor
+
+
```
DanielRosenwasser: Remove extra newline

typescript/libs/service_proxy.py
```diff
@@ -272,3 +272,40 @@ def create_req_dict(self, command_name, args=None):
...
+        return response_dict
+
+
+    def get_errors_at_cursor(self, path, cursor):
+      errors_in_file = self.get_errors_in_file(path)
+      line = cursor[0] + 1
+      column = cursor[1] + 1
+      if errors_in_file['success']:
+          errors_at_cursor = list(filter(lambda error:
+              error['start']['line'] == line  and
```
DanielRosenwasser: stray space

typescript/libs/service_proxy.py
```diff
@@ -272,3 +272,40 @@ def create_req_dict(self, command_name, args=None):
...
+        args = {
+            "file": path
+        }
+        req_dict = self.create_req_dict("semanticDiagnosticsSync", args)
+        json_str = json_helpers.encode(req_dict)
+        response_dict = self.__comm.sendCmdSync(json_str, req_dict["seq"])
+        return response_dict
+
+
+    def get_errors_at_cursor(self, path, cursor):
```
DanielRosenwasser: I don't think `get_errors_at_cursor` should be in `service_proxy.py`. Also, to me it would make more sense to only request fixes at a position if the plugin has already reported a diagnostic on a given region (so you don't have to send a separate diagnostics request). In other words, I'm not sure requesting a code fix should request diagnostics.

Instead, maybe requesting a code fix at a position should check with the global `ProjectErrorListener` and see if it's been `modified`. If so, do nothing, but if not, filter through the errors as you've done here.

Curious to get @billti's thoughts here.

typescript/libs/service_proxy.py
```diff
@@ -272,3 +272,40 @@ def create_req_dict(self, command_name, args=None):
         if args:
             req_dict["arguments"] = args
         return req_dict
+
+    def get_errors_in_file(self, path):
+        args = {
+            "file": path
+        }
+        req_dict = self.create_req_dict("semanticDiagnosticsSync", args)
```
DanielRosenwasser: This would also have to request syntactic diagnostics.

typescript/libs/service_proxy.py
it6: After a bit of exploring 

Current diagnostics only contain `errorMessageString` and `region` leaving out `errorCodes` and `lineOffsets`(which are both required arguments to `getCodeFixes`). I tried passing along `errorCodes` and `offsets` but it didn't come out well with existing code structure.

Definitely agree there is a wasted call to server, but for the amount of refactor  and existing code base constraints it could be a bigger effort

I updated other things

DanielRosenwasser:
Looks sweet! I like how this just applies a quick fix if only one is available. However, I'm getting some strange behavior:

it6:
Which `OS` is this? On `mac` I don't see the `CR` character. Do you have any setting in `sublime` to show line ending characters?  

I looked at `tsc` response `'newText': '  bar: any;\n'` which is expected with `\n` at the end not sure how to debug without reproducing it

it6:
I made a code change to replace `CRLF & CR` characters to `newline` characters, please see if that fixes it

stweedie:
Is there any way this can make it to the released package? It's working well for me when I manually replace the files in the sublime package directory

pjerem:
Since @DanielRosenwasser review has been take into account, what is blocking this PR to be merged ?

orta:
TBH, looks to me like all feedback is addressed and this PR should be merged 👍

PR #681, opened 2018-11-01
Fix region scope

rgant:
Addresses #550. Changes the scope of the error highlights from "keyword" to "invalid.illegal" so themes can control the colors of the highlights.

### DIFF
--- a/typescript/commands/error_list.py
+++ b/typescript/commands/error_list.py
@@ -58,7 +58,7 @@ class TypescriptGoToError(sublime_plugin.TextCommand):
         self.view.add_regions(
             "cur_error",
             [sublime.Region(caret_pos, caret_pos + 1)],
-            "keyword",
+            "invalid.illegal",
             icon,
             sublime.HIDDEN
         )
--- a/typescript/listeners/idle.py
+++ b/typescript/listeners/idle.py
@@ -158,10 +158,10 @@ class IdleListener:
 
         # Highlight error regions in view
         if IS_ST2:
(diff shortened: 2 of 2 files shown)


DanielRosenwasser:
Is there a way I can see this in action? Maybe a screenshot?

Additionally, I tried using scope hunter to check whether scopes were being applied correctly and it doesn't look like they are:

image

rgant:
This is an example of how it looks for me using the theme settings I applied:

screen shot 2018-11-08 at 10 14 17

```
<dict>
	<key>name</key>
	<string>Invalid – Illegal</string>
	<key>scope</key>
	<string>invalid.illegal</string>
	<key>settings</key>
	<dict>
		<key>foreground</key>
		<string>#F92672</string>
	</dict>
</dict>
```

I don't think that scope hunter is actually showing you the TypeScript-Sublime-Plugin added region either. As even without my changes I don't see the scope "keyword" which is used here.

rgant:
I spent some time trying to dig through sublime's api to get the scope of an added region. Short answer is that I don't see a way to do that using the exposed API. The actual implementation of `add_regions` is buryed in the build-in `sublime_api` module, and I'm not prepared to decompile it to try and interpret C source.

But I think it's pretty clear from the Sublime Documentation that `invalid` is the best scope for an error in code: https://www.sublimetext.com/docs/3/scope_naming.html#invalid

DanielRosenwasser:
Seems reasonable. I think we can always revert if it turns out this is too disruptive for some reason.

PR #697, opened 2018-12-31
added ability to pass node_args to forked process

rkoval:
Hello!

I haven't been using this plugin very long because there hasn't been JavaScript support until very recently, but so far it is excellent. However, I was running into issues with tsserver running out of heap memory (I'm using it on a large project). I was able to fix it by just upping the heap size with `--max-old-space-size`, so I wanted to extend the plugin to be able to support any arbitrary arguments to pass to the node process. I believe this PR does it in a pretty succinct and usable way, though please let me know if I should make any tweaks.

Thanks!


roblav96:
@rkoval
You can also set this via an environment variable like so:
```sh
# ~/.bashrc
export NODE_OPTIONS=--max_old_space_size=1024
```

rkoval:
@roblav96 Ooh nice tip! Is there an option within this plugin that will allow that env var to be scoped _only_ to this plugin's node invocation? I don't want that to be globally in my .bashrc such that it applies to every node process that I spawn.

roblav96:
@rkoval When you launch sublime from the terminal, `NODE_OPTIONS=--max_old_space_size=1024 subl -w ~/Project/main.ts`, I certainly wouldn't worry about throwing it in your `.bashrc`, unless your node.js API projects have some major memory leaks, then you've got a whole other problem lol

orta:
@rkoval - this can't be merged until the CLA is accepted ( see https://github.com/microsoft/TypeScript-Sublime-Plugin/pull/697#issuecomment-450613518 )

Otherwise, feels reasonable to me

rkoval:
feel free to close then. i'm not interested in signing the CLA for such a small change.

PR #704, opened 2019-02-24
Fixed node_client.py spamming the console when it does not find node

evandrocoan:
Related to: https://github.com/Microsoft/TypeScript-Sublime-Plugin/pull/648 Stop spamming ERROR: can not send request; node process not


orta:
Any chance you can make this instead only post it once?

PR #707, opened 2019-03-03
Add basic syntax highlighting to signature/quickinfo popups

kbjr:
Adds syntax highlighting to signature popups, based on the currently active theme.

image

Also adds styling to the quickinfo popup

image

### DIFF
--- a/typescript/libs/popup_manager.py
+++ b/typescript/libs/popup_manager.py
@@ -269,7 +269,19 @@ class PopupManager():
                 "index": "{0}/{1}".format(self.signature_index + 1,
                                           len(self.signature_help["items"])),
                 "link": "link",
-                "fontSize": PopupManager.font_size}
+                "fontSize": PopupManager.font_size,
+                "typeColor": theme_styles["type"]["foreground"],
+                "nameColor": theme_styles["name"]["foreground"],
+                "paramColor": theme_styles["param"]["foreground"],
+                "textColor": theme_styles["text"]["foreground"]}
+
+    def get_theme_styles(self):
+        return {
+            "type": self.current_view.style_for_scope("entity.name.type.class.ts"),
+            "name": self.current_view.style_for_scope("entity.name.function"),
+            "param": self.current_view.style_for_scope("variable.language.arguments.ts"),
+            "text": self.current_view.style_for_scope("source.ts")
+        }
 
 _popup_manager = None
 
--- a/typescript/libs/popup_manager.py
+++ b/typescript/libs/popup_manager.py
@@ -202,8 +202,10 @@ class PopupManager():
         def normalize_style(name):
             if name in ['methodName']:
                 return 'name'
-            elif name in ['keyword', 'interfaceName']:
+            elif name in ['interfaceName']:
                 return 'type'
+            elif name in ['keyword']:
+                return 'keyword'
             elif name in ['parameterName', 'propertyName']:
                 return 'param'
             return 'text'
@@ -273,6 +275,7 @@ class PopupManager():
                 "link": "link",
                 "fontSize": PopupManager.font_size,
                 "typeColor": theme_styles["type"]["foreground"],
+                "keywordColor": theme_styles["keyword"]["foreground"],
                 "nameColor": theme_styles["name"]["foreground"],
                 "paramColor": theme_styles["param"]["foreground"],
                 "textColor": theme_styles["text"]["foreground"]}
@@ -280,6 +283,7 @@ class PopupManager():
     def get_theme_styles(self):
         return {
             "type": self.current_view.style_for_scope("entity.name.type.class.ts"),
+            "keyword": self.current_view.style_for_scope("keyword.control.flow.ts"),
             "name": self.current_view.style_for_scope("entity.name.function"),
             "param": self.current_view.style_for_scope("variable.language.arguments.ts"),
             "text": self.current_view.style_for_scope("source.ts")
--- a/typescript/libs/popup_manager.py
+++ b/typescript/libs/popup_manager.py
@@ -202,10 +202,8 @@ class PopupManager():
         def normalize_style(name):
             if name in ['methodName']:
                 return 'name'
-            elif name in ['interfaceName']:
+            elif name in ['keyword', 'interfaceName']:
                 return 'type'
-            elif name in ['keyword']:
-                return 'keyword'
             elif name in ['parameterName', 'propertyName']:
                 return 'param'
             return 'text'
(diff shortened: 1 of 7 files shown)


typescript/libs/popup_manager.py
```diff
@@ -262,14 +262,28 @@ def encode(str, kind):
...
-                "fontSize": PopupManager.font_size}
+                "fontSize": PopupManager.font_size,
+                "typeColor": theme_styles["type"]["foreground"],
+                "nameColor": theme_styles["name"]["foreground"],
+                "paramColor": theme_styles["param"]["foreground"],
+                "textColor": theme_styles["text"]["foreground"]}
+
+    def get_theme_styles(self):
+        return {
+            "type": self.current_view.style_for_scope("entity.name.type.class.ts"),
```
DanielRosenwasser: Kind of unfortunate [we don't do anything for keywords](https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/3f58e5a7a9364e3ad551e7ed4a47a0fd0cb9b608/typescript/libs/popup_manager.py#L206) but I guess it's fine for now. Feel free to send a follow-up.

DanielRosenwasser:
This looks great - I will try to find some time to try the change out locally.

typescript/libs/popup_manager.py
kbjr: I've separated `keyword` chunks out into their own type and added keyword styling to them. I'm not sure what code I can test it on, though, I don't actually see keyword scopes anywhere in signature popups so far just testing out a few different cases.

kbjr:
So, I've also started looking at what it would take to implement this in the other (quick info) popups. It's looking like that will unfortunately be more complicated. Looking at the data that is available when it tries to render, there is nothing broken down into separate tokens like was already done in the signature popups.

```python
{'info_str': '(parameter)&nbsp;foo:&nbsp;Bar', 'doc_str': '', 'error': ''}
```

(from https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/master/typescript/commands/quick_info.py#L67)

Is there some function or such somewhere that can break down the plain text from these info popups into a similar structure as the one used for the signature popups? If that can be done, then these can be highlighted in the exact same way, which would be really nice.

kbjr:
Okay, I've got syntax highlighting working on quickinfo popups now, too.

I need to backport some extra changes I made back into the signature popups now, too (better highlighting pulling in bold/italics as well as color).

I have noticed some of the typing data that tsserver provides is... unusual.

Like calling this a keyword:

image

Or calling the `0 | 1` here a string literal

image

But, for the most part, it works

kbjr:
Any chance of this moving forward? Been sitting idle for a few weeks

mohitsinghs:
@kbjr 2f18af1 looks empty. Isn't it supposed to contain a method named `quick_info_full` in `service_proxy.py` ? I guess you already have those changes locally. BTW, cool work.

kbjr:
Hmm... guess that's what I get for being lazy and copy pasting local changes into github.
I'll go track that down.

kbjr:
Okay, pulled a fresh copy, added the missing method, and confirmed its actually working. Should be good to go.

typescript/libs/service_proxy.py
```diff
@@ -294,6 +294,24 @@ def quick_info(self, path, location=Location(1, 1), on_completed=None):
                 req_dict["seq"]
             )
 
+    def quick_info_full(self, path, location=Location(1, 1), on_completed=None):
```
DanielRosenwasser: Does it make sense to make this `on_completed = lambda: None`?

typescript/libs/popup_manager.py
```diff
@@ -263,13 +263,43 @@ def encode(str, kind):
...
                 "link": "link",
-                "fontSize": PopupManager.font_size}
+                "fontSize": PopupManager.font_size,
+                "typeStyles": theme_styles["type"],
+                "keywordStyles": theme_styles["keyword"],
+                "nameStyles": theme_styles["name"],
+                "paramStyles": theme_styles["param"],
+                "textStyles": theme_styles["text"]}
+
+    def format_css(self, style):
```
DanielRosenwasser: Can you pull these identical helpers into one place?

typescript/libs/service_proxy.py
kbjr: This function is essentially just a copy paste of the non-full version above it. Seemed reasonable to keep the two functions as close to the same as posisble.

typescript/libs/popup_manager.py
kbjr: Sure, I'll look at that when I get home later

kbjr: Okay, I pulled the functions for figuring out the styles out into their own file

typescript/libs/popup_manager.py
```diff
@@ -263,13 +264,20 @@ def encode(str, kind):
         else:
             activeParam = ''
 
+        theme_styles = get_theme_styles(self.view)
```
DanielRosenwasser: I tried this out locally and this errors

```
Traceback (most recent call last):
  File "C:\Users\Daniel\AppData\Roaming\Sublime Text 3\Packages\TypeScriptLang\typescript\libs\work_scheduler.py", line 108, in <lambda>
    sublime.set_timeout(lambda: handler(results), 0)
  File "C:\Users\Daniel\AppData\Roaming\Sublime Text 3\Packages\TypeScriptLang\typescript\libs\popup_manager.py", line 74, in <lambda>
    lambda resp: self.on_response(resp, view))
  File "C:\Users\Daniel\AppData\Roaming\Sublime Text 3\Packages\TypeScriptLang\typescript\libs\popup_manager.py", line 105, in on_response
    self.display()
  File "C:\Users\Daniel\AppData\Roaming\Sublime Text 3\Packages\TypeScriptLang\typescript\libs\popup_manager.py", line 108, in display
    popup_parts = self.get_current_signature_parts()
  File "C:\Users\Daniel\AppData\Roaming\Sublime Text 3\Packages\TypeScriptLang\typescript\libs\popup_manager.py", line 267, in get_current_signature_parts
    theme_styles = get_theme_styles(self.view)
AttributeError: 'PopupManager' object has no attribute 'view'
```

kbjr: Yep, its called different things in the two places its used..
I updated it to correctly reference `self.current_view`

kylebebak:
This would be a really nice improvement. Hope this can get merged soon.

@DanielRosenwasser

DanielRosenwasser:
You're the best @kbjr!

PR #709, opened 2019-03-19
Remove noisy printout in listeners

mykmartin:
This print command pollutes the ST console when editing .ts files. Text operations like un/indent and undo/redo result in printouts of "indent", "undo" etc. in the console.

(I inadvertently deleted the old branch, so this is the same change rebased to the latest head, and with a better branch name.)


orta:
Cool, yeah makes sense, thanks!

mykmartin:
Thanks :)

Issue #489, opened 2016-07-02
How to show red squiggles on errors?


farzher:
I've seen screenshots like this:

![image](https://cloud.githubusercontent.com/assets/1005136/16539399/e054e170-400f-11e6-89c9-e5aeeb850281.png)

I can't figure it out; is that a different plugin? (I tried other plugins but they were buggy, it'd be nice if this plugin could do it)

mhegazy:
What version of ST are you using? i see it working fine on ST3 build: 3114

farzher:
![image](https://cloud.githubusercontent.com/assets/1005136/16549165/e6d48ef0-4169-11e6-87a3-131cda6f9f98.png)

Are you sure it's not from a different plugin? I've reinstalled TypeScript plugin multiple times. Not sure what I could be doing wrong

DanielRosenwasser:
Hey @farzher, have you tried playing around with theme settings to see if that has anything to do with it?

farzher:
It's not my theme's fault because red error lines work when I install ArcticTypescript (but I have other issues with that plugin)

farzher:
@DanielRosenwasser apparently that logic isn't sound, because it is my theme's fault. Other color schemes show the red squiggles. I'll comment again when I figure out how it's my theme's fault, maybe we can add a fix so that it'll work in this case too (apparently ArcticTypescript is able to work fine)

farzher:
This made it invisible:

```
<dict>
	<key>scope</key>
	<string>keyword</string>
	<key>settings</key>
	<dict>
		<key>foreground</key>
		<string>#bB7Dbf</string>
		<key>background</key>
		<string>#9B9FfD10</string>
	</dict>
</dict>
```

(almost completely transparent background on scope `keyword`), maybe the typescript error scope should be changed. I can't figure out how to fix it on my theme without messing up actual keywords.

PR #715, opened 2019-04-28
Changes default max width of HTML popup view from 1500 to 1024; adds error_color and quick_info_popup_max_width settings

kylebebak:
The smallest screen resolution for laptops with a significant number of users worldwide is **1024×768 (xga – standard)**.

This PR lowers the max_width of the quick info popup from 1500 to 1024. Popups with a width of 1500 aren't properly visible on many laptops, including my 13 inch MBP, which has a width of 1280.

It also adds `error_color` and `quick_info_popup_max_width` settings, which can be overridden in `Packages/User/TypeScript.sublime-settings`.

The `error_color` setting would properly solve issues like #489 .

Here's a screenshot to illustrate the popup width issue:

Screen Shot 2019-04-28 at 10 22 08 AM

And here's a screenshot to show type errors underlined in red, using `"error_color": "region.redish"`. The possible error colors are limited by the Sublime Text API. The method used in this PR is the same as the method used by the [Bracket Highlighter plugin](http://facelessuser.github.io/BracketHighlighter/customize/#configuring-highlight-style). See **Clearing Up Color Misconceptions** for a detailed explanation.

Screen Shot 2019-04-30 at 8 07 32 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,15 +85,21 @@ The below features are available via the keyboard shortcuts shown, or via the Co
 |Build		        | (Win)`^B` or `F7`, (OSX) `⌘B` or `F7`   |
 |Error list             | (via Command Palette) |
 
-The "format on key" feature is on by default, which formats the current line after typing `;`, `}` or `enter`.
-To disable it, go to `Preferences` -> `Package Settings` -> `TypeScript` -> `Plugin Settings - User`, and add
-`"typescript_auto_format": false` to the json file.
+The "format on key" feature is disabled by default, which formats the current line after typing `;`, `}` or `enter`.
+To enable it, go to `Preferences` -> `Package Settings` -> `TypeScript` -> `Plugin Settings - User`, and add `"typescript_auto_format": true` to the json file.
 
 For further information about the keyboard shortcuts, please refer to the [`Default.sublime-keymap`](https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/master/Default.sublime-keymap) file for common shortcuts and
 [`Default (OSX).sublime-keymap`](https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/master/Default%20(OSX).sublime-keymap),
 [`Default (Windows).sublime-keymap`](https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/master/Default%20(Windows).sublime-keymap),
 [`Default (Linux).sublime-keymap`](https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/master/Default%20(Linux).sublime-keymap)
 for OS-specific shortcuts.
+
+#### Other settings
+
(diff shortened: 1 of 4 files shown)


kylebebak:
@DanielRosenwasser 

Just bumping and hoping someone can review this and merge it soon. It's nothing special, just some real low-hanging fruit.

Thanks!

DanielRosenwasser:
Hey @kylebebak, looks like there are merge conflicts Can you fix those up?

Also, instead of a fixed width, why not make the tooltips dependent on the viewport/layout width? Maybe just make it the layout width with a margin of 1em (`view.layout_extent()[0] - 2 * view.em_width()`).

See more at https://www.sublimetext.com/docs/3/api_reference.html#sublime.View

kylebebak:
@DanielRosenwasser 
Hey, thanks for the suggestion, this is a great idea and a definite improvement!

I didn't know about this part of ST's API, so thanks for that too.

I'll fix the merge conflicts and push again in a while.

kylebebak:
@DanielRosenwasser 

Hey again! Merge conflicts are now resolved.

Also, a few updates regarding error popup width...

ST has two view methods to get dimensions, `viewport_extent` and `layout_extent`, but neither of these returns the full width of the screen.

The gutters, the sidebar, the file preview column on the right... None of these is included in either of these methods. I looked through the API, and it looks like these are the only methods for getting screen dimensions, so there's no way to use the ST API to get the width of the user's screen.

I understand the reluctance to add settings, but relying on `view.layout_extent` or `view.viewport_extent` means a lot of screen width is unused. On my machine, it's around 850 pixels instead of 1280, about a third of the screen. There are screenshots below showing what it looks like.

I would definitely like the full width of the screen for error popups, because sometimes error messages are long, and the more I can see without scrolling the better. My vote is to keep the `quick_info_popup_max_width` setting. The commit I just pushed falls back to `view.viewport_extent()[0]` if `quick_info_popup_max_width` isn't set. Thanks for that suggestion!


Using `view.layout_extent()[0]`:

Screen Shot 2019-05-23 at 2 27 10 PM


Using `view.viewport_extent()[0]` is a little wider, but still not wide enough for my taste.

Screen Shot 2019-05-23 at 2 26 26 PM

kylebebak:
@DanielRosenwasser 

Is there some reason this can't be merged?

kylebebak:
@DanielRosenwasser 

Just bumping this again. I have another PR that allows users to copy all type and error messages from the tooltip / popup, but I don't want to submit it until this one is closed or merged...

Also, and this is sort of unrelated, but I'd like to volunteer to help maintain this plugin. It could clearly use some help. I've written lots of ST plugins, some with with large code bases and complex behavior, e.g. [Requester](https://requester.org/).

DanielRosenwasser:
Hey @kylebebak, thanks for being patient. I think we'd welcome more contributions, and thinking about it more, 1024 as a default can always be adjusted (or removed altogether).

DanielRosenwasser:
Thanks so much for the PR!

kylebebak:
@DanielRosenwasser 

No, thanks for merging it!

I'll submit the other PR I mentioned (the one that makes it possible to copy info from the type or error popups) in a bit!

PR #727, opened 2019-09-26
#604 - Updated symbols list

### LINKED ISSUE #604: GoTo Definition shows all function calling in file

stweedie:
Fixes #604 where symbol list will show all references to a given function in a file instead of just the function definition.

This is the exact change proposed in #653, but the original author of that change abandoned it.

### DIFF
--- a/Symbols.tmPreferences
+++ b/Symbols.tmPreferences
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+  <key>name</key>
+  <string>Symbol List</string>
+  <key>scope</key>
+  <string>meta.decorator.ts, meta.var.expr.ts, meta.block.ts, meta.field.declaration.ts</string>
+  <key>settings</key>
+  <dict>
+    <key>showInSymbolList</key>
+    <integer>0</integer>
+  </dict>
+</dict>
+</plist>


orta:
Running with this PR gave back zero results for me when I ran with `cmd + r`: 
Screen Shot 2020-01-15 at 11 37 41 AM

( in comparison on master, where it was too much )
( also sorry it took so long to get to this PR )

stweedie:
Thanks for the update @orta. I had to modify the symbols list. Being unable to find proper documentation for how sublime chooses what to put in their symbols list, the best I could do was to have one file to include some symbols (property definitions) and another to exclude other symbols (function calls, for instance).

This will need to be updated to keep up with the tmLanguage file.

Also, one curious thing I noticed is that there is no distinct scope for when you declare a property in the constructor of a class, for instance:

```typescript
export class SomeClass {
  constructor(private someDependency: SomeDependency) { }
}
```

in the above, the 'someDependency' has the scopes: source.ts meta.class.ts meta.method.declaration.ts meta.parameters.ts variable.parameter.ts, none of which describe someDependency as a class parameter. 

and in case it helps, I'm using the following command:
```python
import sublime
import sublime_plugin

class ListScopesCommand(sublime_plugin.TextCommand):
    def run(self, view):
        for reg in self.view.sel():
            print(self.view.scope_name(reg.begin()))
```

if you put that in your packages/User folder, you can set a keybind to call 'list_scopes' and see the scopes of the object currently under your cursor.

orta:
Yeah, it looks like it uses a normal args list for the tokenizer there ( vscode has a very good tokenizer preview BTW, would recommend looking at it in that too ) perhaps it could get added to https://github.com/microsoft/TypeScript-TmLanguage 

I think this is a pretty reasonable approach, it's very likely all of these names aren't changing anytime in the future.

I'm not sure if it's changed anything on my setup though: 

Screen Shot 2020-01-17 at 2 24 07 PM
Screen Shot 2020-01-17 at 2 23 24 PM

I still get a bunch of weird results 

Screen Shot 2020-01-17 at 2 26 20 PM

Screen Shot 2020-01-17 at 2 21 48 PM

Does that feel off to you?

stweedie:
yeah that does look off. Can you find the scopes for those, and I'll try and find some configuration that excludes them while not excluding stuff we would otherwise want

stweedie:
I'm also currently looking at how vscode is implementing their symbols list. I would guess it's not using the text mate scopes as class members defined in the constructor are properly sourced in vscode's symbol list, and there's no difference in scopes between a constructor parameter and a parameter in any class method

orta:
I think vscode uses the tsserver to get the list ([probably `getDocumentHighlights` or `getNavigationBarItems`](https://github.com/microsoft/TypeScript-Website/tree/v2/packages/typescript-vfs#api))

Issue #512, opened 2016-09-06
just installed on sublime text 3, no autocomplete available


BardiaAfshin:
I just installed this package on sublime text 3, no autocomplete is available, restarted ST3, same problem. The file extension is '.ts'

zhengbli:
do you have node in your path?

ZiFFeL1992:
Same here, just formatted my mac and now isn't working (anything, not just autocomplete), I can access node from terminal, and `/usr/local/bin` is in my path... Also use it by `node something`...

I'm working with node 6.5.0 installed from brew.

tavuntu:
same here, the normal autocomplete of Sublime did not work once I installed this package :(
I'm in Manjaro Linux, node 6.10.1 (and I have node in $PATH, yes)

tadeo:
I'm having the same issue here and my productivity is really compromized. Any news about this?

VictorQueiroz:
> The plug-in uses Node.js to run the TypeScript server. The plug-in looks for node in the PATH environment variable (which is inherited from Sublime).
> 
> If the node_path setting is present, this will override the PATH environment variable and the plug-in will use the value of the node_path setting as the node executable to run. See more information in our Tips and Known Issues wiki page.

If you read the documentation, Sublime probablt couldn't find node in your environment, you need to set `NODE_PATH` environment variable.

PR #733, opened 2019-11-10
prevent interruption to tab based autocompletion

matthandlersux:
this PR is a small change that fixes an issue with the expected behavior of sublime text after installing this plugin. 

currently, with the plugin installed, when you type a word fragment and then press `tab`, if typescript does not have any suggestions for autocomplete, then sublime will insert a literal tab/spaces. without the plugin, however, sublime would allow you to tab through several autocomplete suggested words (pulled from matching words in the current document).

the flag that this PR removes (described [here](http://docs.sublimetext.info/en/latest/reference/api.html#sublime_plugin.EventListener.on_query_completions)), makes it so that sublime's standard list of words offered up as part of tab completion are not wiped out. essentially, this restores the expected behavior of typing a word fragment and then hitting `tab` within sublime.

if the reviewers would like, i'd be happy to make this part of the plugin config; it is unclear to me if this flag was set deliberately or not.

interestingly, i think other people previously were trying to raise this same issue, but there was some confusion because of the ambiguity of the phrase "autocomplete does not work". that could mean "when i press tab, the thing i'm used to sublime doing no longer happens", or it could mean "when i try to use autocomplete from this plugin, no results are suggested by tsserver". here are a few where i think this may have been the case: https://github.com/microsoft/TypeScript-Sublime-Plugin/pull/445, https://github.com/microsoft/TypeScript-Sublime-Plugin/issues/442, https://github.com/microsoft/TypeScript-Sublime-Plugin/issues/512

### DIFF
--- a/typescript/listeners/completion.py
+++ b/typescript/listeners/completion.py
@@ -106,7 +106,7 @@ class CompletionEventListener:
             info.last_completion_loc = locations[0]
             self.pending_completions = []
             self.completions_ready = False
-            return completions, sublime.INHIBIT_WORD_COMPLETIONS | sublime.INHIBIT_EXPLICIT_COMPLETIONS
+            return completions, sublime.INHIBIT_EXPLICIT_COMPLETIONS
 
     def handle_completion_info(self, completions_resp):
         """Helper callback when completion info received from server"""


orta:
Perfect, thanks for the write-up, so it looks like this is the before:

Screen Recording 2020-01-15 at 11 42 22 AM 2020-01-15 11_43_18

and after:

Screen Recording 2020-01-15 at 11 41 42 AM 2020-01-15 11_44_07

I can see that is a better experience, let's get this in.

PR #734, opened 2019-11-11
Use `os.path.relpath` on a normalized Packages base path

adill:
The previous method fails because it assumes `packages_path` is
normalized, yet it will not be when Sublime Text is started via the `subl`
command with a non-normalized `PATH` entry
(i.e. `PATH=z:\bin\..\tools\sublime_text`)


orta:
The code looks reasonable to me, and I've tested running it on my computer and popups are working fine 👍

PR #739, opened 2020-03-05
Add setting to manage word completions inhibition

msieurtoph:
#733 created a regression (see my comment : https://github.com/microsoft/TypeScript-Sublime-Plugin/pull/733#commitcomment-37663896)

This PR adds a simple setting key to switch between old and new behavior.

Default behavior is the new one. 
Turning setting `auto_complete_inhibit_word_completions` to True activates the old one.

I hope you will be OK with that.


msieurtoph:
Looking back to my change, I think I found a cleaner way to switch between both behaviors, and properly return api completions only. 
At least, I hope to ...  ;)

The behavior fixed by original PR (#733) is preserved and should still work (as far as I tested it, it worked).

orta:
👍🏻  I'm for this, but it will need documenting in the README first

msieurtoph:
> 👍🏻 I'm for this, but it will need documenting in the README first

Done ;)

orta:
Cool, thanks!

PR #745, opened 2020-05-06
Expose stderr of node process

addreas:
Would be useful to see errors from the node process if it fails to start. For example when editing the typescript_tsdk setting the plugin might silently fail while the node process says that it was unable to find the specified file.


orta:
Thanks!

PR #747, opened 2020-05-09
Add config to exclude name and path suggestions

dekpient:
I was annoyed by the all the irrelevant "warning" suggestions, so I decided to implement a function similar to [`shouldExcludeCompletionEntry` in vscode](https://github.com/microsoft/vscode/blob/master/extensions/typescript-language-features/src/features/completions.ts#L789). The default behaviour stays the same i.e. all suggestions are included.

Please review, cheers.

Related to microsoft/TypeScript-Sublime-Plugin#699 raised by @DanielRosenwasser

### DIFF
--- a/Preferences.sublime-settings
+++ b/Preferences.sublime-settings
@@ -2,5 +2,7 @@
     "typescript_auto_format": false,
     "enable_typescript_language_service": true,
     "enable_language_service_for_javascript": true,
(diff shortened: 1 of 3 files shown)


dekpient:
Hmm, I cannot edit the signed CLA. Could someone revoke it please? I'll sign it again with the correct selection i.e. I'm not making this submission as part of work.

PR #748, opened 2020-05-26
Improve typescript_tsdk lookup

### LINKED ISSUE #538: Typescript-Sublime-Plugin should allow per-project specification of typescript_tsdk

alecmev:
Fix #538

This is flawed, of course, since the server used is never updated once initialized, so gotta restart ST when switching to/from a project with a custom `typescript_tsdk` set, but I can't afford to spend time on a refactor to accommodate this, and this is still much better than the status quo.

### DIFF
--- a/typescript/libs/editor_client.py
+++ b/typescript/libs/editor_client.py
@@ -1,4 +1,7 @@
-﻿from .reference import RefInfo
+﻿import collections
+import logging
+
+from .reference import RefInfo
 from .node_client import ServerClient, WorkerClient
 from .service_proxy import ServiceProxy
 from .logger import log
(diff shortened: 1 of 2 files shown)


typescript/libs/editor_client.py
```diff
@@ -49,19 +52,53 @@ def initialize(self):
...
+                    x = look_for_tsdk(folder)
+                    if x != None:
+                        tsdk_location = x
+                        break
+
+        proc_file = os.path.join(tsdk_location, "tsserver.js")
+        global_vars._tsc_path = os.path.join(tsdk_location, "tsc.js")
+
+        log.warn("Path of tsserver.js: " + proc_file)
+        log.warn("Path of tsc.js: " + get_tsc_path())
```
alecmev, line 101: `log.warn` because now `tsdk_location` deduction is a bit more involved, and it's likely that more people will want to see what's going on. I've considered disabling it for cases where `typescript_tsdk` setting isn't set, but the savings in noise are likely not worth the extra confusion of conditional logging.

alecmev:
@orta @DanielRosenwasser Please, take a look, this fixes the most upvoted open issue.

orta:
Conceptually this looks good, I need to get a working ST3 dev env again to verify this all works before I can merge - I might have time for that tomorrow 👍

alecmev:
@orta Friendly ping :wink:

typescript/libs/editor_client.py
orta: I agree with the log, I agree a bit less with skipping the conditional check, but it's not enough to block the PR 👍

orta:
Thanks for the ping!

typescript/libs/editor_client.py
alecmev: My idea was that people might think that they have a `tsdk_location` in the settings of their current project, while in fact they don't, and the missing log lines could slightly complicate the troubleshooting.

alecmev:
Thank you for dedicating time to this!

orta:
We reverted these changes, after looking at it we realized that the constant lookback for the tsdk could be a cross-user code eval security hole ( `C:/` is writable for all users on a machine )

sahilrajput03:
> We reverted these changes, after looking at it we realized that the constant lookback for the tsdk could be a cross-user code eval security hole ( `C:/` is writable for all users on a machine )

What ..?

image

So how come we'll use the react jsx now, provide help please!!

PR #750, opened 2020-05-29
Add missing required arguments in call to `show_tooltip_popup()`

robsgreen:
When hovering over the error indicator in the gutter, the following error is displayed in the console:
```
Traceback (most recent call last):
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime_plugin.py", line 1082, in run_
    return self.run(edit, **args)
  File "~/Library/Application Support/Sublime Text 3/Packages/TypeScript/typescript/commands/quick_info.py", line 146, in run
    self.show_tooltip_popup(display_point, error_html, None, None)
TypeError: show_tooltip_popup() missing 2 required positional arguments: 'info_html' and 'doc'
```

The call to `show_tooltip_popup()` is missing several required positional arguments, this PR fixes the method call to include what is required.


orta:
Confirmed both the bug, and that this fixes it! Thanks!

PR #752, opened 2020-07-09
Update Yarn 2 instructions

alecmev:
The location where the SDK is put has changed, and there are more `--sdk` commands now, so clarified which one should be used.


orta:
Cool, this seems god.

alecmev:
Thanks!

Issue #765, opened 2020-12-16
TypeScript Sublime master does not seem to work on Windows


orta:
This has been the release blocker.

Running the same project ( https://github.com/danger/danger-js/ ) with a master copy of this plugin does not give hover entry like: 

![Screen Shot 2020-12-16 at 1 09 44 PM](https://user-images.githubusercontent.com/49038/102352694-feac8080-3f9f-11eb-8f4b-0e7fe0adaada.png)

Nor does goto definition work (outside of what I assume is some ST built-in feature.) There are no logs or warnings in the console, and the "TypeScript initializing" message does show.

orta:
I've not been able to figure out how to read any logs for `debug.log` to see whether parts of the code are running. I looked over the diff and none of the new PRs seem to hint at something being platform dependent.

orta:
Confirmed this is fixed with #767

PR #767, opened 2021-01-23
Bypass the TSServer CRLF bug

borodean:
Addresses https://github.com/microsoft/TypeScript-Sublime-Plugin/issues/765.

The plugin broke in the `master` branch after adding commit 64991b18ea7185cd8e1294b19f9cbfc8915a2283. After adding it, when `NodeCommClient` receives a message with an empty header, it tries to read an error from `stderr`.

TSServer doesn't handle line breaks correctly on Windows (https://github.com/Microsoft/TypeScript/issues/3403). Because of this the messages are parsed incorrectly, and the plugin gets into the very piece of code where it now tries to read an error. Before it used to just quit the function at that point and the parsing problem went unnoticed, but now the plugin is stuck trying to read `stderr`.

I found the information about the method of solving the problem here: https://github.com/ycm-core/ycmd/pull/503.

Hopefully, this will unblock the new release.

### DIFF
--- a/typescript/libs/global_vars.py
+++ b/typescript/libs/global_vars.py
@@ -4,6 +4,8 @@ import logging
 import sublime
 from os.path import dirname
 
+IS_WINDOWS = sublime.platform() == "windows"
+
 # determine if the host is sublime text 2
 IS_ST2 = int(sublime.version()) < 3000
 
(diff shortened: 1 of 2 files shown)


orta:
Oh! OK, great - I'll give this a run next week, thanks

orta:
I've given this PR a run locallay and it worked for me on Windows + ST3 - thanks @borodean

